PR #166, opened 2024-10-13
Local file content API & Local file system dialog API

dbolotin:


lib/node/pl-config/src/index.ts
```diff
@@ -4,4 +4,4 @@ export * from './ports';
 export * from './license';
 export * from './auth';
 export * from './storages';
-export * from './package';
+export * from '../../pl-local/src/pl_version';
```
Snyssfx, line 7: Since we already made breaking changes to API in pl-config, this import also should be dropped

lib/node/pl-config/vite.config.mts
```diff
@@ -3,13 +3,12 @@ import path from 'path';
...
-if (!pack['pl-version']) {
-  throw Error('pl-version is required in package.json');
-}
+if (!plVersion) throw Error('pl-version field is required in package.json');
 
 export default PlViteStdNode({
   define: {
-    PL_VERSION: JSON.stringify(pack['pl-version']),
-  },
+    PL_VERSION: JSON.stringify(plVersion)
```
Snyssfx, line 12: This version getter should be dropped, since we moved the version to pl-local

PR #417, opened 2024-12-10
fix: added support PlAgTextAndButtonCell for axis and fixed styles

Gimanh:


sdk/ui-vue/src/components/PlAgDataTable/sources/table-source.ts
```diff
@@ -50,17 +51,35 @@ export const defaultValueFormatter = (value: any) => {
...
     lockPosition: spec.type === 'axis',
     hide: hiddenColIds?.includes(colId) ?? spec.spec.annotations?.['pl7.app/table/visibility'] === 'optional',
     valueFormatter: defaultValueFormatter,
     headerComponent: PlAgColumnHeader,
+    cellRendererSelector: showCellButtonForAxisId
+      ? (params: ICellRendererParams) => {
+          if (spec.type !== 'axis') return;
+
+          const axisId = (params.colDef?.context as PTableColumnSpec)?.id as AxisId;
+          if (axisId?.type === showCellButtonForAxisId.type && axisId?.name === showCellButtonForAxisId.name) {
```
vadimpiven, line 71: axisId could also contain domains and parents, so only correct way to compare is using lodash.isEqual

sdk/ui-vue/src/components/PlAgDataTable/sources/table-source.ts
```diff
@@ -50,17 +51,35 @@ export const defaultValueFormatter = (value: any) => {
...
+            return { component: PlAgTextAndButtonCell };
+          }
+
+          return;
```
vadimpiven, line 75: useless

PR #422, opened 2024-12-11
package-builder, wf: add new sw cmd syntax: expression args

Snyssfx:


tests/package-builder/software-pkg/package.json
```diff
@@ -68,27 +68,27 @@
       "hello": {
         "binary": {
           "artifact": "with-py-environment",
-          "cmd": [ "python", "{pkg}/script1.py", "Hello from python!" ],
+          "command": [ "python", "{pkg}/script1.py", "Hello from python!" ],
```
DenKoren, line 71: Shouldn't it use double brackets?

tests/package-builder/software-pkg/package.json
```diff
@@ -68,27 +68,27 @@
...
           "envVars": [ "MYVAR=1" ]
         }
       },
       "goodbye": {
         "binary": {
           "artifact": "with-py-environment",
-          "cmd": [ "python", "{pkg}/script1.py", "Have a nice day!" ],
+          "command": [ "python", "{pkg}/script1.py", "Have a nice day!" ],
```
DenKoren, line 78: Shouldn't it use double brackets?

tools/package-builder/src/core/package-info.ts
```diff
@@ -680,7 +683,26 @@ const readPackageJson = (filePath: string) => parsePackageJson(fs.readFileSync(f
...
+  }
+
+  const err = JSON.stringify(result.error.format(), null, 2);
+  if (err.includes(`Unrecognized key(s) in object: 'cmd'`)) {
+    throw new Error(`
+    Error: "cmd" was renamed to "command" in package.json in block-software.entrypoints.
+    The syntax has changed: all curly braces should be doubled, for example:
+        {pkg} -> {{pkg}}
+
+    The new "command" field supports rendering of such arguments, for example:
```
DenKoren, line 698: > The new "command" field supports rendering of such arguments, for example:

Let's say not rendering, but clearly point to math expressions and environment variables
> The new "command" field supports math expressions and env variables resolution, for example:

sdk/workflow-tengo/src/exec/descriptor.lib.tengo
```diff
@@ -0,0 +1,92 @@
...
+}
+
+/**
+ * Shows if Platforma Backend is new enough
+ * and sw descriptor is new enough to support expression refs.
+ *
+ * @param descriptor: any - sw descriptor to see if it's new enough.
+ */
+supportExpressions := func(descriptor) {
+	return hasBackendExpressions() && _hasSwExpressions(descriptor)
```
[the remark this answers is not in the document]
Snyssfx: Hmm, I think we should not panic and just return false, it will be a common thing for the next 4 months when Platforma Backend is old, but the block is new and uses expressions. For such cases we still have an old `cmd` in a compiled entrypoint descriptor.

PR #459, opened 2024-12-17
PObjectId schema change

vadimpiven:


lib/node/pl-middle-layer/src/pool/data.ts
```diff
@@ -250,9 +248,10 @@ export function makeDataInfoResource(
...
-export function derivePObjectId(spec: PObjectSpec, data: PlTreeNodeAccessor): PObjectId {
-  const hash = createHash('sha256');
-  hash.update(canonicalize(spec)!);
-  hash.update(String(!isNullResourceId(data.originalId) ? data.originalId : data.id));
-  return hash.digest().toString('hex') as PObjectId;
+export function deriveGlobalPObjectId(blockId: string, exportName: string): PObjectId {
+  return canonicalize({ __isRef: true, blockId, name: exportName } satisfies PlRef)! as PObjectId;
+}
+
+export function deriveLocalPObjectId(outputName: string): PObjectId {
```
vadimpiven, line 255: must also accept platforma tree path. Beware that there are several separate roots, so root id also should be included in path

vadimpiven:
As now PObjectId is stable, we need to derive changes from ResourceInfo.id which is a blobId (and blobId is uniquely derived from blob contents)

dbolotin (approved):

PR #470, opened 2024-12-22
Fixes PTable handle derivation algorithm fix

dbolotin:


lib/node/pl-middle-layer/src/pool/driver.ts
```diff
@@ -396,7 +396,7 @@ function joinEntryToInternal(entry: JoinEntry<PObjectId>): PFrameInternal.JoinEn
 
 function stableKeyFromFullPTableDef(data: FullPTableDef): string {
   const hash = createHash('sha256');
-  hash.update(canonicalize(data.def)!);
+  hash.update(canonicalize(data)!);
```
dbolotin, line 399: @vadimpiven FYI

PR #532, opened 2025-02-03
MILAB-478: add PlSplash page to examples

AndrewKhizhnyak:
MILAB-478: add PlSplash page to UI examples block and improve AgGridVuePage with selector for table state.

PlSplash
image

AgGridVuePage
https://github.com/user-attachments/assets/6cb352ce-0b84-4675-8327-fff09ffa3365

SHM Trees
https://github.com/user-attachments/assets/109892d3-67b7-4866-a8e1-ad3eab506fcc


gramkin:
There are many problems, but first of all: PlSplash is not working properly. 
Let me show you an example:

1) If I use it just as <PlSplash />

image

2) Okay, the developer might think he could use <PlSplash style="height: 100%;" />

image

gramkin (changes requested):
PlSplash needs to be redone

AndrewKhizhnyak:
> PlSplash needs to be redone

Fixed @gramkin

gramkin:
First, we discussed with you and Pavel the need for examples that demonstrate how to use the new PlSplash component in blocks on a separate page within the ui-examples block. I didn’t see these examples implemented. Instead, I only noticed that you added a single PlSplash component to the page. To help illustrate the requirements, I prepared this page for you (see attached screenshot).

The task is as follows:

Reload Page Button:
When clicking the Reload Page button, the page should display a loader with the text "Page loading" centered within the page body (.pl-block-page__body, not the entire html body). During this loading state, users should not be able to interact with any elements on the page except for the header area, which contains the page title and the Reload Page button. This simulates a “loading” state for the page.

Submit Form Button:
When clicking the Submit Form button, a loader should appear over the form along with the button itself. During this loading state, users should not be able to click or focus on any controls on the form.

Reload the Second Paragraph Button:
Similarly, implement a loading state when clicking the Reload the Second Paragraph button.

In all cases, clicking any of these buttons should trigger the respective "loading" state for a couple of seconds before returning to the normal view.

image

gramkin:
@blackcat @AndrewKhizhnyak Let's discuss if necessary

PR #546, opened 2025-02-06
MILAB-1231: workdir limiting integration

DenKoren:
This allows to pause resource propagation to the moment until some other resources become ready


sdk/workflow-tengo/src/exec/exec.tpl.tengo
```diff
@@ -204,7 +229,11 @@ self.body(func(inputs) {
 		"assets,?": { any: validation.resource() },
 		"assetDescriptors,?": { any: {asset: desc.assetScheme } },
 		"assetRules,?": { any: { any: ["string"] } },
-		"assetCache,?": { any: "number" }
+		"assetCache,?": { any: "number" },
+		"quota,?": {
```
Snyssfx, line 233: Could we move it into runOptions? So then no additional value resources will be added to this already big template. `runOptions` is basically 1 JSON value resource

sdk/workflow-tengo/src/exec/index.lib.tengo
```diff
@@ -698,7 +738,11 @@ builder := func() {
 				assets: smart.createMapResource(assetInfos),
 				assetDescriptors: assetDescriptors,
 				assetRules: assetRules,
-				assetCache: assetCache
+				assetCache: assetCache,
+				quota: smart.createJsonResource({
```
Snyssfx, line 742: We might want to add quota only when it was explicitly set, so the old exec templates won't be recalculated because their inputs have changed. What do you think?

sdk/workflow-tengo/src/exec/limits.lib.tengo
```diff
@@ -0,0 +1,149 @@
...
+		 * @return smart.resource - reference to the resulting value resource
+		 */
+		build: func() {
+			ll.assert(queue_name != "", "exec.quota.builder: queue name is mandatory and could not be empty string")
+
+			data := {
+				queue: queue_name
+			}
+			if !is_undefined(cpu_request) {
+				data.cpuCores = cpu_request
```
Snyssfx, line 88: It's not important, but just as a remainder we should use `camelCase` everywhere.

sdk/workflow-tengo/src/units.lib.tengo
```diff
@@ -0,0 +1,13 @@
...
+export ll.toStrict({
+    // Base-2 size modifiers:
+    KiB: 1024,
+    MiB: 1024 * 1024,
+    GiB: 1024 * 1024 * 1024,
+
+    // Decimal modifiers:
+    KB: 1000,
+    MB: 1000,
+    GB: 1000
```
dbolotin, line 12:
```suggestion
    KB: 1000,
    MB: 1000000,
    GB: 1000000000
```

dbolotin (approved):

sdk/workflow-tengo/src/workdir/index.lib.tengo
```diff
@@ -16,9 +18,28 @@ enum := import("enum")
...
+ */
 create := func() {
-	wd := smart.ephemeralBuilder(constants.RTYPE_WORKDIR_CREATE).lockAndBuild()
-	return wd.getField("workdir")
+	// working directories without quotas (old-style)
+	wdCreate := smart.ephemeralBuilder(constants.RTYPE_WORKDIR_CREATE).lockAndBuild()
+	return wdCreate.getField(_WD_CREATE_FIELD_WORKDIR)
+}
+
+createV2 := func(allocation) {
```
Snyssfx, line 38: `allocation` is a reference to the storage space request, right? Let's put a comment here.

Snyssfx (approved):
Looks great, although see my comment about the possible broken recovering of old exec templates

sdk/workflow-tengo/src/units.lib.tengo
DenKoren: Haha. Thanks :)

sdk/workflow-tengo/src/exec/exec.tpl.tengo
DenKoren: Moved it away into a separate service input. It will not affect deduplication at all, because the resources availability is not expected to affect the final execution result, if it succeeds. If it fails, it would be re-run anyway.

sdk/workflow-tengo/src/exec/limits.lib.tengo
DenKoren: Fixed.

sdk/workflow-tengo/src/workdir/index.lib.tengo
DenKoren: Fixed.

sdk/workflow-tengo/src/exec/index.lib.tengo
DenKoren: I put it into separate service field to not affect deduplication at all.

PR #646, opened 2025-03-16
Anchored Column Id

dbolotin:


lib/model/common/src/drivers/pframe/spec/anchored_id.ts
```diff
@@ -0,0 +1,215 @@
...
+
+function domainKey(key: string, value: string): string {
+  return JSON.stringify([key, value]);
+}
+
+/**
+ * Context for resolving and generating anchored references to columns and axes
+ * Maintains maps of known domain values and axes that can be referenced by anchors
+ */
+export class AnchorCtx {
```
Snyssfx, line 22: Ctx is very overloaded word, could we use another word? It might be `GroupedDomains`, not sure.

sdk/workflow-tengo/src/workflow/query-related.tpl.tengo
```diff
@@ -0,0 +1,218 @@
...
+ll := import(":ll")
+slices := import(":slices")
+maps := import(":maps")
+json := import("json")
+bquery := import(":workflow.bquery")
+
+self.awaitState("anchorSpecs", { wildcard: "*" }, "ResourceReady")
+self.awaitState("related", "ResourceReady")
+
+self.body(func(inputs) {
```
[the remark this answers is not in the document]
dbolotin: Please create a ticket for that! 🙏

lib/model/common/src/drivers/pframe/spec/anchored_id.ts
dbolotin: In this case it seems to be a good word. Can't think of a better name.

PR #664, opened 2025-03-24
pfconv exportCsv braking change: for some inputs order of columns in …

vadimpiven:
…exported CSV files could change


dbolotin:
🥳

PR #665, opened 2025-03-25
 pl-errors: add error like shape 

Snyssfx:


lib/model/pl-error-like/src/error_like_shape.ts
```diff
@@ -0,0 +1,63 @@
...
+  message: string;
+  stack?: string;
+  cause?: ErrorLike;
+  errors?: ErrorLike[];
+};
+
+export type ErrorLike = StandardErrorLike | PlErrorLike;
+
+export function ensureErrorLike(error: unknown): ErrorLike {
+  const err = error as any;
```
gramkin, line 24: Why is it "any" there? It's a bit dangerous.

lib/model/common/src/block_state.ts
```diff
@@ -22,7 +22,7 @@ export type BlockState<
    * template */
   ui: UiState;
 
-  /** Outputs rendered with block config */
+  /** Outputs rendered with block confing */
```
gramkin, line 25: Typo

lib/model/pl-error-like/eslint.config.mjs
```diff
@@ -0,0 +1,4 @@
+import { base } from '@milaboratories/eslint-config';
```
gramkin, line 1: Now we have a "common" eslint configuration (for both node and browser universal packages), it is better than the "base" one

lib/model/pl-error-like/src/error_like_shape.ts
gramkin: (btw, we can even use"zod" here)

PR #669, opened 2025-03-25
tengo, pframes: skip exporting files when we have a certain annotation

Snyssfx:


sdk/workflow-tengo/src/pframes/export-pframe.lib.tengo
```diff
@@ -0,0 +1,132 @@
...
+
+	for colName, col in columns {
+		if is_undefined(col.data) {
+			result.add(colName, col.specRaw, undefined)
+			continue
+		}
+
+		shouldHideDataFromUI := skipHideDataFromUI && !is_undefined(col.spec) && col.spec.annotations["pl7.app/hideDataFromUi"] == "true"
+		if shouldHideDataFromUI {
+			// it was called from UI and we shouldn't export this column's data to UI, so we set undefined.
```
DenKoren, line 98:
```suggestion
			// template is called by ML automatic exporter and the column is marked as 'not for UI'. We shouldn't export it.
```

DenKoren (approved):

PR #675, opened 2025-03-28
fix regex to remove pythonMessage line from UI

mirceahasegan:
eslint removes the empty line below pythonMessage which makes the regex fail to remove it.
Fixes boilerplate failure to run block with no python software due to missing model output used in UI.


Snyssfx (approved):
LGTM!

PR #703, opened 2025-04-08
pfconv exportCsv params schema update

vadimpiven:


sdk/workflow-tengo/src/pframes/util.lib.tengo
```diff
@@ -193,6 +193,36 @@ _COLUMN_FILTER_SCHEMA := {
...
+		`type`: `string,regex=column`,
+		// Column id is a column specification filename without extension;
+		// Example: column_id.spec, column_id.datainfo -> column_id
+		`columnId`: `string`
+	},
+	{
+		// Node type discriminator
+		`type`: `string,regex=inner`,
+		// Child nodes to be inner joined
+		`entries`: [_JOIN_ENTRY_SCHEMA]
```
vadimpiven, line 208: Such recursive schema does not compile as `_JOIN_ENTRY_SCHEMA` requires forward declaration. @Snyssfx how can I make this work?

vadimpiven: TypeScript reference https://github.com/milaboratory/pframes/blob/5b046e47ba254739be8d5dbdf06e330da8db9ebb/packages/conv/docs/export_csv.d.ts#L43-L80

PR #732, opened 2025-04-14
Add PlRadio and PlRadioGroup components

kevindetry-milaboratories:


gramkin (changes requested):
all is well, but we need to support any value types, not just strings (by default we use deep equality for comparison)

gramkin:
We discussed and, lest we forget: remove dependency on "name" (and use `isJsonEqual` from `@milaboratories/helpers` package to determine equality between model and value in options), write at least 1 unit test

gramkin (approved):

PR #754, opened 2025-04-24
[uikit] PlAgDataTable: Expose overlay text props

kevindetry-milaboratories:


sdk/ui-vue/src/components/PlAgDataTable/PlAgDataTableV2.vue
```diff
@@ -89,13 +97,25 @@ const props = defineProps<{
...
+   *
+   * If cellButtonInvokeRowsOnDoubleClick = false, the doubleClick event for the row
+   * is not triggered, but will triggered cellButtonClicked event with (key: PTableRowKey) argument.
+   */
   cellButtonInvokeRowsOnDoubleClick?: boolean;
+
+  /**
+   * Override various default labels
+   */
+  labels?: {
```
vadimpiven, line 111: May we should sync this change to https://github.com/milaboratory/platforma/blob/a7821e99773c20a0a2fb3c6316ea55adeec628d3/sdk/ui-vue/src/AgGridVue/useAgGridOptions.ts#L30-L34 so API would be identical?

kevindetry-milaboratories: So, drop the `labels` scope then too, I guess?

PR #770, opened 2025-04-28
MILAB-1946: adopt parser for new line error separator

vgpopov:


lib/node/pl-errors/src/parsed_error.ts
```diff
@@ -247,6 +247,9 @@ export function parseSubErrors(message: string): PlCoreError[] {
       // we need initial stage because apparently the first line
       // of the error doesn't have [I], but is a path line.
       state.stage = 'path';
+    } else if ( line.startsWith('---')) {
```
Snyssfx, line 250: Looks good! Do you think there could be programs that also have lines with `---` though? Maybe we should change it to a more unique set of characters, like 10 random chars, for example

Snyssfx (commented):
LGTM, one question about a string and also we use `changeset` utility in this monorepo, it's a tool for managing monorepos. Try to generate a changeset file from the root of the repository by running `changeset`, and commit this also

lib/node/pl-errors/src/parsed_error.ts
Snyssfx: Or, another variant (since `---` looks good), is to check the separator only when we are parsing `path` part of the message (`state.stage == 'path'`)

Snyssfx (approved):

PR #775, opened 2025-04-29
PlAgDataTable - Skip duplicating label columns

vadimpiven:


sdk/model/src/components/PlDataTable.ts
```diff
@@ -365,47 +366,75 @@ export function getMatchingLabelColumns(
   columns: PColumnIdAndSpec[],
   allLabelColumns: PColumn<TreeNodeAccessor | DataInfo<TreeNodeAccessor>>[],
 ): PColumn<TreeNodeAccessor | DataInfo<TreeNodeAccessor>>[] {
-  const colId = (id: PObjectId, domain?: Record<string, string>) => {
+  // split input columns into label and value columns
+  const inputLabelColumns: typeof columns[number][] = [];
```
kevindetry-milaboratories, line 370: `typeof columns[number][]` seems a little excessive. `typeof columns` or just outright `PColumnIdAndSpec[]` should do the trick

sdk/model/src/components/PlDataTable.ts
```diff
@@ -365,47 +366,75 @@ export function getMatchingLabelColumns(
...
+      const axisId = getAxisId(axis);
+      if (!unlabeledAxes.some((id) => matchAxisId(id, axisId))) {
+        unlabeledAxes.push(axisId);
+      }
+    }
+  }
+
+  // remove axes matched by input label columns
+  for (const labelColumn of inputLabelColumns) {
+    const labelAxisId = getAxisId(labelColumn.spec.axesSpec[0]);
```
kevindetry-milaboratories, line 393: In the previous block you check every element of `axesSpec`, here only the first one. Is this intentional?

sdk/model/src/components/PlDataTable.ts
```diff
@@ -365,47 +366,75 @@ export function getMatchingLabelColumns(
...
-                axesSpec: [{ ...axisId, annotations: labelAxis.annotations }],
-              },
-              data: labelColumn.data,
-            });
-          } else {
-            labelColumns.set(colId(labelColumn.id), labelColumn);
-          }
-        }
+  // search label columns for unmatched axes
+  const labelColumns: typeof allLabelColumns[number][] = [];
```
kevindetry-milaboratories, line 413: Same thing about the types: `typeof allLabelColumns` should probably be just as fine.

sdk/model/src/components/PlDataTable.ts
vadimpiven: fixed, thanks!

sdk/model/src/components/PlDataTable.ts
vadimpiven: yes, label column has only one axis, check isLabelColumn predicate

PR #781, opened 2025-04-30
check network: storages check; drivers: download: range query initial

Snyssfx:


lib/node/pl-middle-layer/src/network_check/network_check.ts
```diff
@@ -140,6 +151,18 @@ export async function checkNetwork(
...
     pythonSoftwareCheck: await pythonSoftware(client, 'Jack'),
+    storageToDownloadReport: await downloadFromEveryStorage(
+      logger,
+      client,
+      lsDriver,
+      downloadClient,
+      {
+        bytesLimit: ops.everyStorageBytesLimit,
+        minFileSize: ops.everyStorageMinFileSize,
+        maxFileSize: ops.everyStorageMaxFileSize,
```
DenKoren, line 162: If you download small part of file using range query, you actually can choose any file, even terabytes-big.
I think we should make this upper border unlimited, because logic can fail in storage that has only big files (big raw datasets only)

DenKoren (approved):

lib/node/pl-drivers/src/drivers/ls.ts
```diff
@@ -42,8 +42,19 @@ export interface InternalLsDriver extends sdk.LsDriver {
    * To be used in tests and in implementation of the native file selection UI API.
    * */
   getLocalFileHandle(localPath: string): Promise<sdk.LocalImportFileHandle>;
+
+  listRemoteFilesWithSizes(storage: sdk.StorageHandle, fullPath: string): Promise<ListRemoteFilesResultWithAdditionalInfo>;
```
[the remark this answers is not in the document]
Snyssfx: I'm not sure about adding sizes, since it's a heavy operation for virtual storages we have locally (1 `stat` syscall for every file we have).

lib/node/pl-middle-layer/src/network_check/network_check.ts
Snyssfx: Ok, let's remove it

PR #787, opened 2025-05-06
MILAB-2402: bin path depends on os

vgpopov:
[MILAB-2402](https://www.notion.so/mixcr/backend-runenv-python-python-venv-is-not-available-on-windows-1d83a83ff4af80b29ad1e7bf0282ffff?pvs=4)


lib/node/pl-deployments/src/local/pl.ts
```diff
@@ -159,8 +165,10 @@ export async function localPlatformaInit(logger: MiLogger, _ops: LocalPlOptions)
 
     const configPath = upath.join(workDir, LocalConfigYaml);
 
-    logger.info(`writing configuration '${configPath}'...`);
-    await fsp.writeFile(configPath, ops.config);
+    if (!ops.useCustomConfig) {
```
DenKoren, line 168: Why not writing file when it does not exist? This will simplify process of hacking the instance: you run it for the first time to generate the config, modify it and restart the app to use altered version. You need to re-generate it - drop the file and start the backend so it generates you new one.

For me `useCustomConfig` opt name is felt like you can put a path to your own configuration file and work with it. Better to name it closer to the meaning, like `doNotGenerateConfig` or `doNotRegenerateConfig`.

DenKoren (approved):

PR #801, opened 2025-05-12
MILAB-2402: add local python scripts to pythonpath

vgpopov:


DenKoren:
Was merged into main inside some other branch after rebase. I see PYTHONPATH and changeset inside main. This is because of 'revert for revert for revert, after merge' and that kind of stuff.

PR #802, opened 2025-05-12
drivers: download: add a cache by range

Snyssfx:
- test coverage in pl-drivers
- getContent in download blob driver uses ranges
- these methods are updated in middle-layer everywhere.
- Tests


lib/node/pl-middle-layer/src/network_check/network_check.test.ts
```diff
@@ -8,7 +8,7 @@ import { testCredentials } from './test_utils';
 // - locally
 // - stop minio and retest locally
 // - on demo server
-test('should return network connectivity report for a platform endpoint set in PL_ADDRESS env', async () => {
+test.skip('should return network connectivity report for a platform endpoint set in PL_ADDRESS env', async () => {
```
DenKoren, line 11: Why?

sdk/model/src/config/model.ts
```diff
@@ -149,26 +149,36 @@ export type CfgResourceValueAsJson = {
 
 export type CfgBlobContent = {
   type: 'GetBlobContent';
+  fromBytes: number | undefined;
```
DenKoren, line 152: Why not to put it into its own type (i.e. `rangeContent`) and use composition here and in all types below?

DenKoren (approved):

lib/node/pl-middle-layer/src/network_check/network_check.test.ts
Snyssfx: It's 80 seconds test with no asserts, it's just to see what is the output of the network check tool. It should be skipped, however I forgot to do it in the previous PR.

lib/node/pl-drivers/src/drivers/download_blob/download_blob.test.ts
```diff
@@ -53,50 +44,56 @@ test('should download a blob and read its content', async () => {
...
+});
 
-test('should not redownload a blob a file already exists', async () => {
+test('should download a blob range and read its content with a range', async () => {
   await TestHelpers.withTempRoot(async (client) => {
-    const logger = new ConsoleLoggerAdapter();
-    const dir = await fsp.mkdtemp(path.join(os.tmpdir(), 'test-download-1-'));
+    const dir = await fsp.mkdtemp(path.join(os.tmpdir(), 'test-download-driver'));
+    const driver = await genDriver(client, dir, genSigner(), { cacheSoftSizeBytes: 700 * 1024, nConcurrentDownloads: 10 });
+    const downloadable = await makeDownloadableBlobFromAssets(client, fileName);
```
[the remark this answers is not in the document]
Snyssfx: I'm not sure it's correct, I have 2 or 3 tests that downloads only a part of the file via range. There I get either `4`, `42`, `2\n` or other parts of the whole `42\n` file.

lib/model/common/src/drivers/blob.ts
```diff
@@ -19,6 +19,35 @@ export interface BlobHandleAndSize<
...
+/** Range in bytes, from should be less or equal than to. */
+export type RangeBytes = {
+  /** Included left border. */
+  from: number;
+
+  /** Excluded right border. */
+  to: number;
+};
+
+export function newRangeBytesOpt(from?: number, to?: number): RangeBytes | undefined {
```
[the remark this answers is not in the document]
Snyssfx: Yeah, I checked, Vadim said it's ok

lib/node/pl-drivers/src/drivers/download_blob/download_blob.test.ts
```diff
@@ -54,44 +45,50 @@ test('should download a blob and read its content', async () => {
...
+    expect(blob).toBeUndefined();
+
+    console.log(`should download a blob range: awaiting change`)
+    await c.awaitChange();
+
+    console.log(`should download a blob range: getting the blob second time`)
+    const blob2 = await c.getValue();
+    expect(blob2).toBeDefined();
+    expect(blob2!.size).toBe(3);
+    expect((await driver.getContent(blob2!.handle, { from: 0, to: 2 }))?.toString()).toBe('42');
```
Snyssfx, line 69: Here's the test, for example

PR #838, opened 2025-05-26
Accept undefined as QuickJS function param

vadimpiven:


lib/node/pl-middle-layer/src/js_render/context.ts
```diff
@@ -186,6 +196,7 @@ export class JsExecutionContext {
   }
 
   public importObjectUniversal(handle: QuickJSHandle): unknown {
+    if (handle === undefined) return undefined;
```
gramkin, line 199: This is not type safe, is it?

PR #841, opened 2025-05-27
linker columns for createPFrameForGraphs

erohinaelena:


sdk/model/src/components/PFrameForGraphs.ts
```diff
@@ -216,10 +298,15 @@ export function createPFrameForGraphs<A, U>(
     .entries.map((v) => v.obj)
     .filter(isPColumn);
 
-  const columnsWithCompatibleFromUpstream = enrichColumnsWithCompatible(blockColumns, upstreamColumns);
+  const allAvailableColumns = [...blockColumns, ...upstreamColumns];
+  // all linker columns always go to pFrame - even it's impossible to use some of them they all are hidden
+  const linkerColumns = allAvailableColumns.filter((item) => item.spec.annotations?.[IS_LINKER_COLUMN] === 'true');
```
vadimpiven, line 303: please make a helper like https://github.com/milaboratory/platforma/blob/4bbb60d825f820546c0ad3a7098cbb2bf0455061/sdk/model/src/components/PlDataTable.ts#L356

sdk/model/src/components/PFrameForGraphs.ts
```diff
@@ -49,24 +50,90 @@ function getKeysCombinations(idsLists: AxisId[][]) {
   return result;
 }
 
+type LinkerColumnsMap = Record<string, Record<string, { id: string, spec: PColumnSpec }>>;
```
vadimpiven, line 53: we have type `PColumnIdAndSpec`, please use it instead of `{ id: string, spec: PColumnSpec }`

PR #861, opened 2025-06-02
MILAB-2846: setup http public settings for storages in builtin mod

vgpopov:
Add new settings to the backend configuration to allow using blocks with assets in the backend:
- add HTTP public endpoint settings


lib/node/pl-config/src/common/storages.ts
```diff
@@ -106,7 +113,7 @@ export async function createDefaultLocalStorages(workdir: string): Promise<Stora
   return storages;
 }
 
-export function newDefaultLocalStorages(workdir: string): StoragesSettings {
+function newDefaultLocalStorages(workdir: string, externalURL: string): StoragesSettings {
```
Snyssfx, line 116: Here we could just require externalURL as `string`, not as `?: string`, right? And then we won't need the runtime check below. Also, in JS world it's `externalUrl`, we don't capitalize abbreviations here (strange world 🤷🏼‍♂️)

Snyssfx (approved):

Snyssfx (dismissed):
Wait, we also need to bump pl version here:
https://github.com/milaboratory/platforma/blob/main/lib/node/pl-deployments/package.json#L4

Snyssfx (approved):

PR #867, opened 2025-06-04
MILAB-2783: drivers: download: sparse files cache

Snyssfx:
- add a tracer bullet test
- add a sparse cache for ranges for `getContent` function of `downlaod_blob` driver
- as the configuration of download driver has changed, I updated all ml funcs that works with it
- updated download-file block to see if everything is working

### DIFF
--- a/lib/node/pl-drivers/src/drivers/download_blob/download_blob.ts
+++ b/lib/node/pl-drivers/src/drivers/download_blob/download_blob.ts
@@ -70,9 +70,8 @@ export type DownloadDriverOps = {
   cacheSoftSizeBytes: number;
 
   /**
-   * A soft limit of the amount of sparse cache, in bytes.
-   * Once exceeded, the download driver will start deleting blobs one by one
-   * when they become unneeded.
+   * A hard limit of the amount of sparse cache, in bytes.
+   * Once exceeded, the download driver will start deleting blobs one by one.
    *
    * The sparse cache is used to store ranges of blobs.
    * */
@@ -136,7 +135,7 @@ export class DownloadDriver implements BlobDriver {
     ops: DownloadDriverOps,
   ): Promise<DownloadDriver> {
     const driver = new DownloadDriver(logger, clientDownload, clientLogs, saveDir, rangesCacheDir, signer, ops);
-    await driver.rangesCache.initFromDir(driver.rangesCacheDir);
+    await driver.rangesCache.reset();
 
     return driver;
   }
--- a/lib/node/pl-drivers/src/drivers/download_blob/sparse_cache/cache.ts
+++ b/lib/node/pl-drivers/src/drivers/download_blob/sparse_cache/cache.ts
@@ -2,8 +2,12 @@ import { RangeBytes } from '@milaboratories/pl-model-common';
 import { ensureDirExists, fileExists, mapEntries, MiLogger } from '@milaboratories/ts-helpers';
 import { promises as fs } from 'node:fs';
 import path from 'node:path';
-import { addRange, existRange, Ranges, rangesFileName, rangesFilePostfix, rangesSize, readRangesFile, writeRangesFile } from './ranges';
+import { addRange, doesRangeExist, Ranges, rangesFileName, rangesFilePostfix, rangesSize, readRangesFile, writeRangesFile } from './ranges';
 import { writeToSparseFile } from './file';
+import { functions } from '@milaboratories/helpers';
+
+/** The implementer of SparseCacheRanges could throw it if ranges were corrupted. */
+export class CorruptedRangesError extends Error {}
 
 /** Extracted ranges methods to be able to store ranges somewhere else (e.g. in memory for tests). */
 export interface SparseCacheRanges {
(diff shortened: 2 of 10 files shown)


lib/node/pl-drivers/src/drivers/download_blob/sparse_cache/ranges.ts
```diff
@@ -0,0 +1,81 @@
...
+  return fPath + rangesFilePostfix;
+}
+
+export async function readRangesFile(logger: MiLogger, path: string): Promise<Ranges> {
+  let ranges: Ranges = { ranges: [] };
+  try {
+    const file = await fs.readFile(path, 'utf8');
+    ranges = Ranges.parse(JSON.parse(file));
+  } catch (e: unknown) {
+    if (!(e instanceof Error && 'code' in e && e.code === 'ENOENT')) {
```
vgpopov, line 27: I don't agree with throwing an exception. If the file was corrupted, we shouldn't show an unfixable error to the user. Could we try loading it again from the beginning?

lib/node/pl-drivers/src/drivers/download_blob/sparse_cache/ranges.ts
```diff
@@ -0,0 +1,81 @@
...
+      throw e;
+    }
+    // If the file does not exist, assume the ranges are empty.
+  }
+
+  return normalizeRanges(ranges);
+}
+
+export async function writeRangesFile(path: string, ranges: Ranges) {
```
vgpopov, line 36: Does user not open platforma twice? It's little bit scared that we don't open file with exclusive lock.

And do we always all storage from the same thread that excludes concurrent file access? (I think yes, but could tell pls more about constraints)

lib/node/pl-drivers/src/drivers/download_blob/sparse_cache/cache.ts
```diff
@@ -0,0 +1,153 @@
...
+    await this.ensureCleared();
+  }
+
+  async setNoClear(key: string, range: RangeBytes, data: Uint8Array): Promise<void> {
+    this.keyToLastAccessTime.set(key, new Date());
+
+    const ranges = await this.ranges.get(key);
+    this.size -= rangesSize(ranges);
+
+    await this.storage.write(key, data, range.from);
```
DenKoren, line 127: I'd set a safety check here to ensure you really write the data covered by range. When you write MORE than range it is safe. When you write less - it means next time you will think you wrote the data you really didn't.

Let's put sanity check here that data size is `==` to the range size (or at least `>=`).

lib/node/pl-drivers/src/drivers/download_blob/sparse_cache/cache.ts
```diff
@@ -0,0 +1,153 @@
...
+
+    return undefined;
+  }
+
+  async set(key: string, range: RangeBytes, data: Uint8Array): Promise<void> {
+    await this.setNoClear(key, range, data);
+    await this.ensureCleared();
+  }
+
+  async setNoClear(key: string, range: RangeBytes, data: Uint8Array): Promise<void> {
```
DenKoren, line 121: The name confuses a little. What 'NoClear' means?

lib/node/pl-drivers/src/drivers/download_blob/sparse_cache/ranges.test.ts
```diff
@@ -0,0 +1,57 @@
...
+    },
+    {
+      name: 'inner range',
+      input: { ranges: [{ from: 0, to: 20 }, { from: 5, to: 15 }] },
+      expected: { ranges: [{ from: 0, to: 20 }] },
+    },
+    {
+      name: 'inner range with outer range',
+      input: { ranges: [{ from: 5, to: 15 }, { from: 20, to: 30 }, { from: 0, to: 20 }] },
+      expected: { ranges: [{ from: 0, to: 30 }] },
```
DenKoren, line 44: Please, add here at least one test where you have to leave more than one range in the result.
Otherwise, if you write simple algo that 'merges all ranges from lowest to largest', it will pass this testset. Test we keep the gap between.

lib/node/pl-drivers/src/drivers/download_blob/download_blob.ts
```diff
@@ -67,6 +68,16 @@ export type DownloadDriverOps = {
...
   cacheSoftSizeBytes: number;
+
+  /**
+   * A soft limit of the amount of sparse cache, in bytes.
+   * Once exceeded, the download driver will start deleting blobs one by one
+   * when they become unneeded.
+   *
+   * The sparse cache is used to store ranges of blobs.
+   * */
+  rangesCacheMaxSizeBytes: number;
```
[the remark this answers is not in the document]
Snyssfx: Yep, this is a hard limit. We don't manage links in this cache, since it works directly from UI or from pframes avoiding computables (and hence reference counting) logic.

lib/node/pl-middle-layer/src/middle_layer/ops.ts
```diff
@@ -124,10 +127,11 @@ export const DefaultDriverKitOpsSettings: Pick<
   logger: new ConsoleLoggerAdapter(),
   blobDriverOps: {
     cacheSoftSizeBytes: 8 * 1024 * 1024 * 1024, // 8 GB
+    rangesCacheMaxSizeBytes: 8 * 1024 * 1024 * 1024, // 8 GB
     nConcurrentDownloads: 10,
   },
   downloadBlobToURLDriverOps: {
-    cacheSoftSizeBytes: 100 * 1024 * 1024, // 100MB
+    cacheSoftSizeBytes: 1 * 1024 * 1024 * 1024, // 1 GB
```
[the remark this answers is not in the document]
Snyssfx: Yeah, we could override any of the middle layer settings from desktop, so it's a matter of implementing settings on the desktop side for these. I'll create a ticket.

lib/node/pl-drivers/src/drivers/download_blob/download_blob.ts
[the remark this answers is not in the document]
DenKoren: So, what's the difference with `cacheSoftSizeBytes`? The description in comment is pretty the same for both.

DenKoren (approved):

PR #869, opened 2025-06-05
Feat: update vite version

AStaroverov:


etc/ui-vue-playground/vite.config.ts
```diff
@@ -4,11 +4,4 @@ import vue from '@vitejs/plugin-vue';
 // https://vitejs.dev/config/
 export default defineConfig({
   plugins: [vue()],
-  css: {
```
AStaroverov, line 7: already default behaviour for vite 6

PR #872, opened 2025-06-06
MILAB-2910: Standartize the error property across all uikit form components

AStaroverov:


lib/ui/uikit/src/composition/useErrorMessage.ts
```diff
@@ -0,0 +1,27 @@
+import { isErrorLike, tryDo } from '@milaboratories/helpers';
+
+// order of errors is important, we will show first
+export function useErrorMessage(...errors: unknown[]): undefined | string {
```
kevindetry-milaboratories, line 4: I think there's a convention in Vue to name [composables](https://vuejs.org/guide/reusability/composables) as `useSomething`. This doesn't seem to be a composable?

gramkin (changes requested):

PR #879, opened 2025-06-10
MILAB-1924: wf: more notifications, compression, exec ctx, smart API

Snyssfx:


sdk/workflow-tengo/src/ll.lib.tengo
```diff
@@ -287,12 +287,130 @@ parseUrl := func(url) {
...
+ * It's a map for cases when
+ * we need to have a place for global variables.
+ * For example, tengo modules instantiate on each import and don't store global state.
+ * So we can use this map to store temporary data.
+ *
+ * Pros: it can store non-serializable data, like methods.
+ * Cons: it's not saved after the script execution.
+ */
+getExecutionCtx := func() {
+	return plapi.getExecutionCtx()
```
DenKoren, line 319: Lost assertion. As long as you do it for other new methods.

sdk/workflow-tengo/src/smart.lib.tengo
```diff
@@ -390,7 +397,7 @@ resource = func(r) {
 		 * Resource is ready when all inputs are set to the terminal values.
 		 */
 		isFinal: func() {
-			return self.info().ResourceReady || self.info().IsDuplicate || self.info().HasErrors
+			return self.info().IsFinal
```
DenKoren, line 400: It's not backward-compatible, right? I thought we need `if feats.is...` here.

sdk/workflow-tengo/src/smart.lib.tengo
```diff
@@ -699,13 +706,25 @@ resource = func(r) {
...
 		},
 
+		/**
+		 * Gets a flag from key value storage associated with resource
+		 *
+		 * @param key: string - key
+		 * @return bool
+		 */
+		kvGetFlag: func(key) {
+			return tx.resourceKeyValueGetFlag(self.id, key)
```
DenKoren, line 725: Let's do assertions with `feats...` here and in all other new methods. At least we'll get clear message on why we broke instead of 'indefined is not callable'

DenKoren (dismissed):

PR #884, opened 2025-06-11
[ml/desktop] v2api Standard Error serialization

gramkin:


lib/model/common/src/errors.ts
```diff
@@ -0,0 +1,135 @@
...
+
+  return {
+    name: error.name,
+    message: error.message,
+    stack: error.stack,
+    ...(error.cause !== undefined && { cause }),
+  };
+}
+
+export type ResultOrError<S, F = Error> = {
```
AStaroverov, line 80: already exist vore

lib/model/common/src/errors.ts
```diff
@@ -0,0 +1,135 @@
...
+};
+
+export function unwrapResult<T>(result: ResultOrError<T>): T {
+  if (result.error) {
+    throw result.error;
+  }
+  return result.value;
+}
+
+export function serializeResult<T>(result: ResultOrError<T>): ResultOrError<T, SerializedError> {
```
AStaroverov, line 95: can you tell why we need Maybe abstraction?

lib/model/common/src/errors.ts
```diff
@@ -0,0 +1,135 @@
...
+  }
+
+  return String(`Non-Error value (${typeof value}) was thrown: ${value}`);
+}
+
+export function ensureError(value: unknown): Error {
+  if (value instanceof Error) {
+    return value;
+  }
+
```
AStaroverov, line 45: if (isErrorLke(value)) {
const err = new Error(value.message);
err.name = value.name;
err.stack = value.stack;
err.cause = value.cause;
}

dbolotin:
bugbot run

dbolotin:
bugbot run

dbolotin (approved):

lib/model/common/src/errors.ts
AStaroverov: you can replace this in better place

> And we lost the error type

next string `return new Error(JSON.stringify())`, you also lost type of error type

PR #885, opened 2025-06-12
MILAB-2820: use FS support for new ssh depolyment

vgpopov:


DenKoren (dismissed):

Snyssfx (dismissed):
Everything is cool and it sure does read better now 👍🏼 

I suggest 3 things:
1. chore: we could apply eslint that will autofix trailing whitespaces and formatting: `eslint --fix` in `pl-deployments` folder (and you could setup the linter in the editor).
2.  suggestion: we could reorder functions to match the order of execution in `platformaInit`. Like we have `doStepStop` at the top, but it should be somewhere in the middle.
3. issue: also please see cursor's suggestions, the first one is important I think

Snyssfx (approved):

PR #893, opened 2025-06-16
fix: print valid env variable name in error message when no storage url

DenKoren:


.changeset/clever-carrots-kick.md
```diff
@@ -0,0 +1,5 @@
+---
+'@platforma-sdk/package-builder': patch
+---
+
+Print valid env variable name in error message about absent storage URL
```
vgpopov, line 5: Could you tell pls why we should do it in middle layer and why we're talking about storage URL? It's not a new feature, right? Is it other storage?

PR #918, opened 2025-06-25
Fix for closed channel error. All gRPC clients are now used via GrpcClientProvider

dbolotin:


lib/node/pl-client/src/core/ll_client.ts
```diff
@@ -125,17 +146,52 @@ export class LLPlClient {
...
-    return this._grpcTransport;
+  private providerCleanupCounter = 0;
+
+  /**
+   * Creates a provider for a grpc client. Returned provider will create fresh client whenever the underlying transport is reset.
+   *
+   * @param clientConstructor - a factory function that creates a grpc client
+   */
+  public createGrpcClientProvider<Client>(clientConstructor: (transport: GrpcTransport) => Client): GrpcClientProvider<Client> {
+    // We need to cleanup providers periodically to avoid memory leaks.
```
vgpopov, line 173: Why do memory leaks occur?

PR #932, opened 2025-06-30
Introduced metaExtra to the process column

dbolotin:


sdk/workflow-tengo/src/tpl/base.lib.tengo
```diff
@@ -181,10 +181,18 @@ tpl := func(ops) {
 		},
 
 		/**
-		 * @return map[string]field: - a map of inputs to the fields of the input map.
+		 * @return map[string]field: - a map of meta inputs to the fields of the input map.
+		 * @deprecated Use rawMetaInputs instead.
```
DenKoren, line 185: Better to put deprecation warnings to the first line of the comment. Otherwise it is easier to miss it when reading.

DenKoren (approved):

dbolotin:
bugbot run

PR #937, opened 2025-07-01
Final fixes for resource allocation in workflow SDK

dbolotin:


sdk/workflow-tengo/src/workdir/index.lib.tengo
```diff
@@ -128,6 +118,10 @@ _fill := func() {
 	values := {}
 	dirs := {}
 
+	queue := undefined
+	cpu := undefined
+	ram := undefined
```
DenKoren, line 123: Shouldn't we set default values here? At least for queue.
Or add assertion into build() function that queue name is not undefined.
CPU and RAM are OK to be empty, queue is mandatory.

sdk/workflow-tengo/src/workdir/index.lib.tengo
```diff
@@ -269,13 +303,41 @@ _fill := func() {
...
+					quotaBuilder.queue(execConstants.BATCH_QUEUE)
+				} else {
+					quotaBuilder.queue(queue)
+				}
+
+				storageSpaceRef := limits.storageSpaceRequestBuilder().
+					quota(quotaBuilder.build()).
+					build()
+
+				allocation = storageSpaceRef.allocation
```
DenKoren, line 336: After brief discussion with @vgpopov , we decided that it is better to name all references with Ref suffix to be clear for the reader.

When you see `allocation` you can think of real allocation resource to be there. When you see `allocationRef` it is clear for you that there might be unresolved reference under the hood.

DenKoren (approved):

sdk/workflow-tengo/src/workdir/index.lib.tengo
dbolotin: Ok, good point. Though there are many places where similar naming would be useful, this is one out of hundreds of similar places.

sdk/workflow-tengo/src/workdir/index.lib.tengo
DenKoren: It seems you already had to update workdir test in neighbour package because of this new restriction. I still think we need at least validation in `build()` with clear error message.

PR #961, opened 2025-07-10
MILAB-2445: ssh: add logs, add dev container

Snyssfx:


lib/node/pl-deployments/src/ssh/supervisord.ts
```diff
@@ -115,9 +115,16 @@ autostart=true
...
+stdout_logfile=${remoteWorkDir}/platforma_stdout.log
+stdout_logfile_maxbytes=10000
+stdout_logfile_backups=5
+stderr_logfile=${remoteWorkDir}/platforma_stderr.log
+stderr_logfile_maxbytes=10000
+stdout_logfile_backups=5
 `;
 }
 
+/** @deprecated, we don't use minio anymore, use generation of the config above. */
```
vgpopov, line 127: Unfortunately, we do for the old installations. That's why I didn't mark the method as deprecated

vgpopov (approved):

lib/node/pl-deployments/src/ssh/supervisord.ts
Snyssfx: Yeah, it's fine for deprecated methods to be used imho, or else we'd delete them from the code

PR #969, opened 2025-07-14
MILAB-3235: createPFrameForGraphs update

erohinaelena:


sdk/model/src/components/PFrameForGraphs.ts
```diff
@@ -203,6 +203,10 @@ function getAdditionalColumnsForColumn(
   return [column, ...additionalColumns];
 }
 
+function isColumnNotReady(c: PColumn<PColumnDataUniversal>) {
+  return c.data instanceof TreeNodeAccessor && !c.data.getIsReadyOrError();
```
vadimpiven, line 207: this check is not correct, PColumnDataUniversal can also hold `DataInfo<TreeNodeAccessor>`, and in such case you have to check multiple TreeNodeAccessors for the column. And I think that better check would be `isColumnReady`, negation could easily be added at the place of call

sdk/model/src/components/PFrameForGraphs.ts
```diff
@@ -219,8 +223,31 @@ export function createPFrameForGraphs<A, U>(
...
   blockColumns: PColumn<PColumnDataUniversal>[] | undefined,
 ): PFrameHandle | undefined {
-  if (!blockColumns) return undefined;
+  // if current block doesn't produce own columns then use all columns from result pool
+  if (!blockColumns) {
+    const columns = new PColumnCollection();
+    columns.addColumnProvider(ctx.resultPool);
+
+    const allColumns = columns.getColumns(() => true, { dontWaitAllData: true, overrideLabelAnnotation: false }) ?? [];
+    const allAxes = new Map<CanonicalizedJson<AxisId>, AxisId>();
```
vadimpiven, line 232: do you know that there is a functional constructor for the map? I think for the case of initialization using it looks much better and also provides automatic type inference:
```
const allAxes = new Map(
  allColumns
    .flatMap((column) => column.spec.axesSpec)
    .map((axisSpec) => {
      const axisId = getAxisId(axisSpec);
      return [canonicalizeJson(axisId), axisId];
    })
);
```

sdk/model/src/components/PFrameForGraphs.ts
```diff
@@ -219,8 +223,31 @@ export function createPFrameForGraphs<A, U>(
...
+        const aid = getAxisId(id);
+        allAxes.set(canonicalizeJson(aid), aid);
+      }
+    }
+
+    // additional columns are duplicates with extra fields in domains for compatibility if there are ones with partial match
+    const extendedColumns = enrichCompatible(allAxes, allColumns);
+
+    // if at least one column is not yet ready, we can't show the table
+    if (extendedColumns.some(isColumnNotReady))
```
vadimpiven, line 244: I think we can check just after getting all colums, to skip additional calculations when they are not needed

sdk/model/src/components/PFrameForGraphs.ts
```diff
@@ -203,6 +205,20 @@ function getAdditionalColumnsForColumn(
   return [column, ...additionalColumns];
 }
 
+function isColumnReady(c: PColumn<PColumnDataUniversal>) {
+  let ready = true;
+  if (c.data instanceof TreeNodeAccessor) {
+    ready = c.data.getIsReadyOrError();
+  } else if (isDataInfo(c.data)) {
+    visitDataInfo(c.data, (v) => {
+      if (!v.getIsReadyOrError()) {
```
vadimpiven, line 214: may be `ready = ready && v.getIsReadyOrError()` would be even more readable?
Please also add `export` to funciton. And add helper
```
export function allColumnsReady(columns: PColumn<PColumnDataUniversal>[]): boolean {
  return columns.every(isColumnReady);
}
```
I think that what should be used in most places and in your code. Just write `if !(allColumnsReady(...)) return undefined;`

sdk/model/src/components/PFrameForGraphs.ts
```diff
@@ -203,6 +205,24 @@ function getAdditionalColumnsForColumn(
   return [column, ...additionalColumns];
 }
 
+export function isColumnReady(c: PColumn<PColumnDataUniversal>) {
+  let ready = true;
+  if (c.data instanceof TreeNodeAccessor) {
+    ready = ready && c.data.getIsReadyOrError();
+  } else if (isDataInfo(c.data)) {
+    visitDataInfo(c.data, (v) => {
+      if (ready) {
```
vadimpiven, line 214: тут if не нужен, само присвоение `ready = ready && v.getIsReadyOrError();` уже его содержит

vadimpiven (approved):

PR #1036, opened 2025-08-06
do not require package.sw.json to exist in bin package archives for backend before 1.42

DenKoren:
This also fixes MILAB-3574 (ptabler dependency install issues broke tests in CI)

### DIFF
--- a/tests/workflow-tengo/src/exec/test/package.test.ts
+++ b/tests/workflow-tengo/src/exec/test/package.test.ts
@@ -5,19 +5,15 @@ tplTest.concurrent('package-loads-and-installs', async ({ helper, expect }) => {
   const result = await helper.renderTemplate(
     false,
     'exec.test.pkg.pkg-install',
-    ['installed', 'descriptor'],
+    ['installed'],
     (tx) => ({}),
   );
   const installedOutput = result.computeOutput('installed', (a) => a?.getDataAsJson());
-  const descriptorOutput = result.computeOutput('descriptor', (a) => a?.getDataAsJson());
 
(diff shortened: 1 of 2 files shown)


sdk/workflow-tengo/src/exec/package.lib.tengo
```diff
@@ -29,7 +29,15 @@ DEFAULT_ALLOCATION_RAM := "1GiB"
 DEFAULT_ALLOCATION_CPU := 1
 
 get := func(registryName, pkgName) {
-	res := smart.structBuilder(_RTYPE_PACKAGE_GET)
+	res := smart.structBuilder(_RTYPE_PACKAGE_GET).lockAndBuild()
+
+	// TODO: remove this hack and all mentions of META_KEY_DESCRIPTOR after 2026-02-01
```
vgpopov, line 34: nit: I suggest thinking about the approach with the date. When I face a similar comment, I always change a date or leave code without changes because I don't understand why we can remove it after some date. Writing the condition for removing code will be more useful and clearer.

DenKoren: The only condition is date here: we need a time gap between releasing new version of backend and thinking it is widely used enough to drop obsolete code support.

Why you feel you need to change the date in comment? Does it affect anything :)

vgpopov: No, no, it's not a date. The trigger is the new backend usage, and the date is a sign. But if I don't know about the trigger and don't trust the sign. That is an issue. I should spend some time investigating to find the root trigger.

PR #1045, opened 2025-08-08
Fix: ts-builder correct resolve path to libs

AStaroverov:


tools/ts-builder/src/commands/utils/command-runner.ts
```diff
@@ -2,6 +2,7 @@ import { spawn } from 'child_process';
 
 export function runCommand(command: string, args: string[]): Promise<void> {
   return new Promise((resolve, reject) => {
+    console.log(`↳ ${command} ${args.join(' ')}`);
```
vadimpiven, line 5: intentional or forgot?

vadimpiven (approved):

PR #1047, opened 2025-08-11
MILAB-3704: force stop pl process if it rejects to stop within timeout

DenKoren:
When backend process started for builtin mode does not stop within timeout of 10 seconds - we just leave attempts to kill the process and then fail to start backend, showing the user a message about OS permissions.

Instead, we should do these things:
- set stop timeout on backend side to some controlled value (now backend has default of 30s when ML waits for 10s and constantly fails)
- make ML to control timout and force process to stop afterwards.
- to not kill unexpected process because of mistake - do simple process name validation for safety.

### DIFF
--- a/.changeset/funny-regions-agree.md
+++ b/.changeset/funny-regions-agree.md
@@ -1,0 +1,5 @@
+---
+'@milaboratories/pl-deployments': patch
+---
+
+Force stop backend process if it exceeeds timeout after gentle stop signal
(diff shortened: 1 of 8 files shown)


DenKoren:
> Are you sure that timeout is useful only for builtin mode and it's not necessary for ssh?

SSH uses supervisor which kills backend on its own following its internal logic.
I think we do not need to alter that part of logic.

PR #1048, opened 2025-08-11
MILAB-3136: support python docker container

rfiskov:


tools/package-builder/assets/python-dockerfile.template
```diff
@@ -0,0 +1,16 @@
+# Auto-generated Dockerfile for Python package
+FROM python:${PYTHON_VERSION}
+
+WORKDIR /app
+
+# Install dependencies to leverage Docker cache
+${PYTHON_INSTALL_DEPS}
```
vgpopov, line 7: Could we use sh condition and move condition to docker template? It will be more clear and readable.
And why do leave this comment? Do you use some additional cache strategies`RUN--mount=type=cache,target=/root/.cache`?

tools/package-builder/assets/python-dockerfile.template
```diff
@@ -0,0 +1,16 @@
...
+
+# Copy package source
+COPY . /app/
+
+# Set Python path
+ENV PYTHONPATH=/app
+
+# Default command (will be overridden by entrypoint)
```
vgpopov, line 15: You don't specify entrypoint. Could you plesase add  more clear comment why we do what we do because it's totally different from custom docker image approach

tools/package-builder/src/commands/publish/all.ts
```diff
@@ -43,6 +43,8 @@ export default class PublishAll extends Command {
 
     core.publishDescriptors();
 
-    // TODO: don't forget to add docker here, when we support it
+    core.publishDockerImages({
```
vgpopov, line 46: there was a bug here? Docker isn't publish in publishPacakges?

rfiskov: Yes, indeed, nothing had been sent in this command earlier.

rfiskov:
Screenshot 2025-08-21 at 18 07 15

DenKoren (approved):

PR #1090, opened 2025-08-20
wc -l for tengo SDK

dbolotin:


lib/ptexter/package.json
```diff
@@ -34,6 +34,24 @@
             "{pkg}/phead-lines.py"
           ]
         }
+      },
+      "wc-l": {
+        "binary": {
+          "artifact": {
```
DenKoren, line 40: What do you think of creating single package with single requirements.txt, but different entrypoints for it?

PR #1104, opened 2025-08-27
MILAB-2939 : adapt full testset for google cloud batch

rfiskov:


rfiskov:
Example of a local run:
Screenshot 2025-08-26 at 11 31 56

tests/workflow-tengo/vitest.config.mts
```diff
@@ -3,7 +3,7 @@ import { defineConfig } from 'vitest/config';
 export default defineConfig({
   test: {
     watch: false,
-    testTimeout: 15000,
+    testTimeout: 40000,
```
DenKoren, line 6: Do not understand the need to update this global default timeout.
As far as I understood from other file changes, we still have default timeouts for each test that requires longer than this 15 seconds, unless TEST_TIMEOUT env variable is set. This means, 15 000 here will leave logic more-or-less the same as before your PR.
Let's leave it 15, to not make tests that do not cause real executions take longer before failing in case of a problem.

Am I miss something?

DenKoren (approved):

tests/workflow-tengo/vitest.config.mts
rfiskov: Yes - with this PR the “long” default is driven solely by the TEST_TIMEOUT env var, so setting a global default here is redundant. Per-test overrides stay as they are; if TEST_TIMEOUT is set (e.g., in CI) it applies globally, and if it isn’t, tests rely on their own local timeouts.

PR #1105, opened 2025-08-27
feat: add tool for update dependencies in workspace

vgpopov:


pnpm-workspace.yaml
```diff
@@ -1,125 +1,126 @@
...
-  - tests/package-builder/10.asset
-  - tests/package-builder/20.runenv
-  - tests/package-builder/30.software-pkg
-  - tests/package-builder/90.catalogue
-
-  # Etc
-
-  - etc/uikit-playground
-  - etc/ui-vue-docs
+- lib/util/helpers
```
DenKoren, line 4: Was it really required? To unindent all records in list.

tools/blocks-deps-updater/.gitignore
```diff
@@ -0,0 +1,3 @@
+pkg-*.tgz
+pkg-*.zip
+package.sw.json
```
DenKoren, line 3: Drop it. We don't generate it any more.
```suggestion
```

DenKoren (approved):

PR #1106, opened 2025-08-28
MILAB-3859: real content addressable

DenKoren:


tools/package-builder/src/commands/build/all.ts
```diff
@@ -54,14 +55,22 @@ export default class BuildAll extends Command {
         skipIfEmpty: flags['package-id'] ? false : true, // do not skip 'non-binary' packages if their IDs were set as args
       });
 
-      if (!flags['no-docker-push']) {
+      if (!flags['skip-docker-build'] && !flags['skip-docker-push'] && core.buildMode === 'release') {
```
vgpopov, line 61: We should add workround for monorepo because ptexter and so on trying to push at local build wihtout creds.

tools/package-builder/src/core/core.ts
```diff
@@ -278,75 +284,74 @@ export class Core {
       throw new Error(`Context '${context}' not found`);
     }
 
-    const tag = docker.tagFromPackage(this.pkg.packageRoot, buildParams);
+    const localTag = `pl.pkg.${pkg.name}.${util.currentArch()}:local`;
```
vgpopov, line 289: please, move to docker file

tools/package-builder/src/core/core.ts
```diff
@@ -278,75 +284,74 @@ export class Core {
...
-      entrypoint: "${entrypoint.join(' ')}"
+    this.logger.info(`Building docker image...`);
+    this.logger.debug(`dockerfile: '${dockerfile}'
+context: '${context}'
+localTag: '${localTag}'
+entrypoint: '${entrypoint.join('\', \'')}'
     `);
 
-    docker.build(context, dockerfile, tag);
+    docker.build(context, dockerfile, localTag);
```
vgpopov, line 298: Functions below looks make the same thing please move them to one `build` function

tools/package-builder/src/core/core.ts
```diff
@@ -278,75 +284,74 @@ export class Core {
...
-    const args = ['publish'];
-    if (options?.npmPublishArgs) {
-      args.push(...options.npmPublishArgs);
-    }
+    writeBuiltArtifactInfo(artInfoPath, {
+      type: pkg.type,
+      platform: util.joinPlatform(os, arch),
+      registryURL: pkg.registry.downloadURL,
+      registryName: pkg.registry.name,
+      pathForSwJson: pkg.namePattern,
```
vgpopov, line 350: What means pathForSwJson = namePattern? Don't get it =(  
Maybe namePatter isn't pattern?

tools/package-builder/src/core/docker.ts
```diff
@@ -20,13 +17,9 @@ export function entrypointNameToOrigin(name: string): string {
   return name.substring(0, suffixIndex);
 }
 
-export function localImageExists(tag: string): boolean {
-  const result = spawnSync('docker', ['image', 'ls', '--filter=reference=' + tag, '--format=\'{{.ID}}\''], {
+export function getImageID(tag: string): string {
+  const result = spawnSync('docker', ['image', 'ls', '--filter=reference=' + tag, '--format={{.ID}}'], {
     stdio: 'pipe',
-    env: {
-      ...process.env, // Inherit all environment variables from the parent process
```
vgpopov, line 27: it's needed for authorization if i recall correctly

tools/package-builder/src/core/core.ts
vgpopov: I guess that we can face a case when there are a few build in the same runner. Could we use old tag version for local tags?

tools/package-builder/src/core/core.ts
```diff
@@ -181,6 +181,10 @@ export class Core {
     for (const pkgID of packagesToBuild) {
       const pkg = this.getPackage(pkgID);
 
+      if (pkg.type === 'docker') {
```
vgpopov, line 186: why you want skip docker build in CI?

tools/package-builder/src/core/core.ts
DenKoren: In sw.json, the location of software/docker is stored in `package` field.
For crossplatform software it keeps direct archive path
For binary software that is platform-dependent (star, python/java/r interpreters, cellranger and so on) it keeps placeholders (`tra-la-la-{os}-{arch}.tgz`)

When we learn to use content-addressable archive paths in registies, we will put direct path here. But not now :(

tools/package-builder/src/core/docker.ts
DenKoren: `docker image ls` does not request remote registry.
AFAIR, docker uses file stored in user's home to keep auth tokens.

PR #1157, opened 2025-09-18
MILAB-4102: support custom environment vars in run environments

DenKoren:


tests/package-builder/20.runenv/package.json
```diff
@@ -6,16 +6,20 @@
...
   "block-software": {
     "entrypoints": {
       "python": {
         "environment": {
           "artifact": {
-            "registry": "example",
+            "registry": "example-python-registry",
             "type": "environment",
             "runtime": "python",
+            "python-version": "3.12.10",
```
vgpopov, line 19: Why did we add env version here. There are two place with version: in dependencies and here. What will happen if they are different?

DenKoren: I want to make build scripts that prepare run environment to use this version instead of version specifier in build script positional parameter. Like we already did for R.
Now we already have version number in three places: in package name, in entrypoint name and in `build.js <version>` argument that builds python in CI.

Python docker preparation magic uses entrypoint name to detect python version to use in docker.
We already have place where versions may diverge: if build script inside runenv NPM package gets started as `build.js 3.12.10`, but entrypoint name is `runenv-python-3:3.10.0`, package-builder will try to use python 3.10.0 as base docker image.

This is an effort to have single place of truth for two points: runenv build script and package-builder.

vgpopov: It's difficult to understand what is going on, even with your comment.

DenKoren: 'python-version' is version of python interpreter.
All the rest are versions of packages (binary package, npm package)

PR #1182, opened 2025-10-03
MILAB-4125: move annotations to common source

erohinaelena:


lib/model/common/src/drivers/pframe/spec/spec.ts
```diff
@@ -179,14 +205,38 @@ export type Annotation = Metadata & Partial<{
 // >>;
 
 export type AnnotationJson = MetadataJson<Annotation>;
+
+const ValueTypeSchema = z.union([
```
vadimpiven, line 209: z.enum(['Int', 'Long', 'Float', 'Double', 'String'] as const)

lib/model/common/src/drivers/pframe/spec/spec.ts
```diff
@@ -142,16 +156,28 @@ export const Annotation = {
 
 export type Annotation = Metadata & Partial<{
   [Annotation.Alphabet]: 'nucleotide' | 'aminoacid' | string;
+  [Annotation.AxisNature]: 'homogeneous' | 'heterogeneous' | 'scaleCompatible';
```
vadimpiven, line 159: Are you sure there would be no new variant in here? If there could be a new variant added at some point, please add `| string` in the end, it will force to add default case to all switches

PR #1476, opened 2026-03-02
fix: improve file upload resilience

dbolotin:
## Summary

- Treat S3 `ExpiredToken` errors as recoverable during file uploads. When STS session credentials expire during long multipart uploads (800+ chunks), S3 returns HTTP 400 with `ExpiredToken`. Previously classified as non-recoverable `BadRequestError`, permanently killing the upload. Now classified as recoverable `NetworkError`, allowing the retry loop to request a fresh pre-signed URL.

## Context

Large file uploads that exceed the STS token lifetime (1-12h depending on AWS config) would fail terminally at an arbitrary chunk with no way to recover. The fix parses the S3 error XML body to distinguish `ExpiredToken` from other 400 errors.

## Test plan

- [ ] Upload a large file (800+ chunks) and verify upload completes successfully
- [ ] Verify that genuine 400 errors (malformed requests) still fail as non-recoverable
- [ ] Verify retry behavior: after `ExpiredToken`, the next attempt gets a fresh pre-signed URL from the backend

### DIFF
--- a/lib/node/pl-drivers/src/clients/upload.ts
+++ b/lib/node/pl-drivers/src/clients/upload.ts
@@ -316,24 +316,33 @@ async function checkExpectedMTime(path: string, expectedMTimeUnix: bigint) {
   }
 }
 
+function isExpiredTokenError(body: string): boolean {
+  return body.includes("<Code>ExpiredToken</Code>");
+}
+
 function checkStatusCodeOk(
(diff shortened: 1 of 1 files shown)


DenKoren (approved):

dbolotin:
Fixed in #1533

PR #1518, opened 2026-03-18
MCP server for Platforma Desktop

blackcat:
## Summary

- MCP server package (`lib/node/pl-mcp-server/`) with Streamable HTTP transport
- Project CRUD tools (list, create, open, close, delete)
- Block management tools (add, remove, run, stop)
- Block state tools (get overview, get/set block data)
- Await tools (await_block_done with two-phase wait)
- Log tools (get_block_logs from block outputs, get_app_log for Electron log)
- Screenshot and UI interaction tools (click, type, scroll, press_key, execute_js)
- Block navigation (select_block, list_available_blocks)
- Multi-session support (per-client MCP server instances)
- Lifecycle callbacks for desktop UI sync

## Test plan

- [ ] Integration tests: 16 tests covering project CRUD, block management, state read/write, pipeline await, timeout
- [ ] Manual desktop testing: VDJ 003 Tiny Trees project setup via MCP
- [ ] Screenshot verification of all UI operations
- [ ] Block log reading from MiXCR outputs

This PR introduces a new `@milaboratories/pl-mcp-server` package that exposes Platforma Desktop's project/block management capabilities as an MCP (Model Context Protocol) server over Streamable HTTP transport, enabling AI agents to drive the desktop application programmatically. The implementation is well-structured — tools are cleanly separated by domain, the multi-session transport lifecycle is correctly handled (including the race-condition guard added in the latest commit), and the test suite covers the key CRUD/pipeline/state paths with proper setup/teardown.

A few issues were found:

- **`data-query.ts` — dead pFrame detection condition**: `lowerPath.endsWith(".pFrame")` is applied after `toLowerCase()`, so the uppercase `F` ensures the condition is always `false`. Harmless today because `lowerPath.includes("pframe")` already covers the common case, but the dead branch is misleading.
- **`data-query.ts` — O(n²) `vectorToJson` in `query_table`**: `vectorToJson` is called once per cell (rows × columns) rather than once per column. At the 1000-row limit with a wide table this materialises the full column array for every single cell, which is wasteful and should be fixed before large tables are queried.
- **`logs.ts` — silent data loss when `shouldUpdateHandle` is `true`**: Live log handles can signal that the handle is stale via `shouldUpdateHandle`. The current code silently omits the entry from results rather than still decoding `response.data` or surfacing a staleness hint, making log reads silently return nothing for live handles.
- **`data-query.ts` — identical ternary branches**: Two locations use `s.type === "column" ? s.spec.name : s.spec.name`, which is a no-op ternary likely left from a copy-paste.
- **`types.ts` — `getOpenedProject` typed as `Promise<any>`**: This disables TypeScript's checks on all opened-project method calls across the block/state/log/await tools.
</details>

- Safe to merge for initial functionality, but the O(n²) vectorToJson and silent log data loss should be addressed before heavy use against wide or live-log tables.
- The server transport lifecycle and project/block CRUD tools are solid. The three correctness issues in data-query.ts and logs.ts are real bugs (not just style) — one causes silent data loss for live log handles, one causes quadratic work at the documented 1000-row limit, and one is a dead detection condition — but none affect the core project management or block execution paths covered by the existing tests.
- `lib/node/pl-mcp-server/src/tools/data-query.ts` and `lib/node/pl-mcp-server/src/tools/logs.ts` need the most attention before these tools are used against real data.
</details>

| Filename | Overview |
|----------|----------|
| lib/node/pl-mcp-server/src/server.ts | Core HTTP server with multi-session transport management. The race condition fix (closed-flag guard before `transports.set`) and top-level error handler are solid. Port retry logic correctly preserves the request handler across recreated servers. |
| lib/node/pl-mcp-server/src/tools/data-query.ts | Three issues: (1) dead condition `lowerPath.endsWith(".pFrame")` never fires because the path is already lowercased; (2) `vectorToJson` is called once per cell rather than once per column causing O(n²) work at max 1000-row limit; (3) two ternary branches with identical `s.spec.name` expressions indicate a likely copy-paste error. |
| lib/node/pl-mcp-server/src/tools/logs.ts | Silent data loss: when `shouldUpdateHandle` is `true`, log data available in `response.data` is discarded and the key is silently omitted from results, giving callers no indication that logs exist but the handle is stale. |
| lib/node/pl-mcp-server/src/tools/await.ts | Two-phase await (overview poll → stable block state) is well-structured. The `Promise.race` + `AbortSignal.timeout` + `setTimeout` triple is slightly redundant but functionally correct. |
| lib/node/pl-mcp-server/src/tools/types.ts | `getOpenedProject` is typed as `Promise<any>`, disabling type-checking across all block/state/log/await tools that consume the returned object. |
| tests/mcp-server/src/with-mcp.ts | Clean test harness wrapping MiddleLayer + PlMcpServer lifecycle. Proper cleanup in finally block. `getFreePort` correctly exported for direct-HTTP tests in server.test.ts. |

</details>

</details>

<details open><summary><h3>Comments Outside Diff (3)</h3></summary>

1. `lib/node/pl-mcp-server/src/server.ts`, line 830-832 ([link](https://github.com/milaboratory/platforma/blob/78c1e6e7aa27deccef452f3abdb209c219450af3/lib/node/pl-mcp-server/src/server.ts#L830-L832)) 

   <a href="#">P1</a> **`delete_project` uses `entry.id` instead of `entry.rid`**

   All other project operations (`open_project`, `close_project`) call `ml.openProject(entry.rid)` / `ml.closeProject(entry.rid)` using the `ResourceId`. Only `delete_project` calls `ml.deleteProject(entry.id)`. If `entry.id` is a different type (e.g., a plain string database identifier rather than the `ResourceId`), this will pass the wrong argument to `deleteProject` and likely fail at runtime or silently delete nothing. Compare with the pattern used everywhere else:

   

   `````markdown
   Path: lib/node/pl-mcp-server/src/server.ts
   Line: 830-832

   Comment:
   **`delete_project` uses `entry.id` instead of `entry.rid`**

   All other project operations (`open_project`, `close_project`) call `ml.openProject(entry.rid)` / `ml.closeProject(entry.rid)` using the `ResourceId`. Only `delete_project` calls `ml.deleteProject(entry.id)`. If `entry.id` is a different type (e.g., a plain string database identifier rather than the `ResourceId`), this will pass the wrong argument to `deleteProject` and likely fail at runtime or silently delete nothing. Compare with the pattern used everywhere else:

   

   `````
   </details>

2. `lib/node/pl-mcp-server/src/server.ts`, line 971-973 ([link](https://github.com/milaboratory/platforma/blob/78c1e6e7aa27deccef452f3abdb209c219450af3/lib/node/pl-mcp-server/src/server.ts#L971-L973)) 

   <a href="#">P2</a> **Missing `refreshState` before awaiting overview**

   `list_projects` and `resolveProject` both call `ml.projectList.refreshState()` before awaiting the value, ensuring fresh data. `get_project_overview` directly calls `project.overview.awaitStableValue()` without first triggering a refresh. This may return a stale cached value, especially immediately after mutations (e.g., `set_block_data`).

   Consider adding a refresh step:

   

   `````markdown
   Path: lib/node/pl-mcp-server/src/server.ts
   Line: 971-973

   Comment:
   **Missing `refreshState` before awaiting overview**

   `list_projects` and `resolveProject` both call `ml.projectList.refreshState()` before awaiting the value, ensuring fresh data. `get_project_overview` directly calls `project.overview.awaitStableValue()` without first triggering a refresh. This may return a stale cached value, especially immediately after mutations (e.g., `set_block_data`).

   Consider adding a refresh step:

   

   `````
   </details>

3. `lib/node/pl-mcp-server/src/server.ts`, line 1009-1060 ([link](https://github.com/milaboratory/platforma/blob/78c1e6e7aa27deccef452f3abdb209c219450af3/lib/node/pl-mcp-server/src/server.ts#L1009-L1060)) 

   <a href="#">P2</a> **`await_block_done` does not handle all terminal error states**

   The polling loop only checks for `"Done"` and `"Limbo"` as terminal states. If there are other terminal/error states in the Platforma calculation status enum (e.g., `"RunningError"`, `"Failed"`, or any future additions), the loop will keep polling until the full `timeout` elapses rather than returning immediately with the error condition.

   This is a usability issue — an AI agent waiting for a failed block will silently wait the full timeout (default 2 minutes) instead of being told immediately that the block failed. Consider exhaustively listing all non-`Running` states that should short-circuit the loop, or inverting the check to only continue looping on explicitly `"Running"` / `"NotCalculated"` states.

   `````markdown
   Path: lib/node/pl-mcp-server/src/server.ts
   Line: 1009-1060

   Comment:
   **`await_block_done` does not handle all terminal error states**

   The polling loop only checks for `"Done"` and `"Limbo"` as terminal states. If there are other terminal/error states in the Platforma calculation status enum (e.g., `"RunningError"`, `"Failed"`, or any future additions), the loop will keep polling until the full `timeout` elapses rather than returning immediately with the error condition.

   This is a usability issue — an AI agent waiting for a failed block will silently wait the full timeout (default 2 minutes) instead of being told immediately that the block failed. Consider exhaustively listing all non-`Running` states that should short-circuit the loop, or inverting the check to only continue looping on explicitly `"Running"` / `"NotCalculated"` states.

   `````
   </details>
</details>

`````markdown
Path: lib/node/pl-mcp-server/src/tools/data-query.ts
Line: 30-32

Comment:
**Dead condition — `.endsWith(".pFrame")` never matches after `toLowerCase()`**

`lowerPath` is already fully lowercased, so `lowerPath.endsWith(".pFrame")` (capital `F`) will always evaluate to `false`. Any handle located at a path that would match only the second condition (but not the `includes("pframe")` first condition) will be silently misclassified as a pTable or not detected at all.

```suggestion
      if (lowerPath.includes("pframe") || lowerPath.endsWith(".pframe")) {
```
(or simply remove the dead second condition, since `lowerPath.includes("pframe")` already subsumes `.endsWith(".pframe")`)

---

Path: lib/node/pl-mcp-server/src/tools/data-query.ts
Line: 235-244

Comment:
**O(n²) performance: `vectorToJson` called once per cell instead of once per column**

`vectorToJson(vectors[c], actualRows)` is invoked inside the inner loop — once for every row `r` and every column `c`. Each call allocates and fills an array of `actualRows` elements, only to immediately index into it with `[r]`. For the maximum `limit` of 1000 rows and e.g. 20 columns, this means 20,000 `vectorToJson` invocations each scanning 1000 entries = 20 million iterations of the bit-mask logic instead of 20,000.

Materialise each column vector once outside the row loop:

```typescript
// Pre-materialise all column vectors once
const columnVectors = vectors.map((v) => vectorToJson(v, actualRows));

const rows: unknown[][] = [];
for (let r = 0; r < actualRows; r++) {
  const row: unknown[] = columnVectors.map((col) => col[r]);
  rows.push(row);
}
```

---

Path: lib/node/pl-mcp-server/src/tools/logs.ts
Line: 46-51

Comment:
**Silent data loss when `shouldUpdateHandle` is `true`**

When `shouldUpdateHandle` is `true` (the driver signalling that the handle is stale and a fresh one is available in `response.newHandle`), the log content is available in `response.data` but the code silently discards it and writes nothing into `results[key]`. The caller receives an empty object `{}` with no indication that logs exist but the handle was just stale.

At a minimum, fall through to decode the data even when the handle should be refreshed, or surface a placeholder that tells the caller the handle needs updating:

```typescript
const response = await logDriver.lastLines(handle, lines);
results[key] = new TextDecoder().decode(response.data);
// Optionally surface handle staleness so callers can refresh
if (response.shouldUpdateHandle) {
  results[`${key}:handleStale`] = true as unknown as string;
}
```

---

Path: lib/node/pl-mcp-server/src/tools/data-query.ts
Line: 161-162

Comment:
**Identical ternary branches — `name` field is always `s.spec.name`**

Both branches of the conditional are identical, so the ternary is dead code. This pattern appears in two places (`get_block_outputs` and `query_table`). The likely intent was to access a different property for the axis-key spec type (e.g. `s.spec.columnId` or similar).

```suggestion
              name: s.spec.name,
```

The same issue exists at the `query_table` column-headers section (line ~271).

---

Path: lib/node/pl-mcp-server/src/tools/types.ts
Line: 14

Comment:
**`getOpenedProject` return type is `any` — loses type safety**

`getOpenedProject` is typed as returning `Promise<any>`, which propagates `any` into every tool that calls it (all block/state/log/await tools). This disables TypeScript's checks on the project object's methods and properties (e.g. `project.mutateBlockStorage`, `project.getBlockState`, `project.overview`, etc.).

Consider typing it as the concrete opened-project type exported by `pl-middle-layer`:

```typescript
getOpenedProject: (projectId: string) => Promise<OpenedProject>;
```

`````

</details>

<sub>Reviews (2): Last reviewed commit: ["fix: address PR review comments"](https://github.com/milaboratory/platforma/commit/6d5fc521c2dd9986ebfeb1b6b861f8ffae22ce39) | [Re-trigger Greptile](https://app.greptile.com/api/retrigger?id=25597476)</sub>

### CODE REVIEW

File: lib/node/pl-mcp-server/src/tools/block-state.ts [hunk 1]
```diff
@@ -0,0 +1,87 @@
...
+      const project = await ctx.getOpenedProject(projectId);
+      await project.mutateBlockStorage(
+        blockId,
+        { operation: "update-block-data", value: data },
+        ctx.getAuthorMarker(),
+      );
+      return textResult({ ok: true });
+    },
+  );
+}
```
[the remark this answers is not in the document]
dbolotin: It is even more important as this specific endpoint may expose user-defined functions, and doing this might be even simpler as what we've discussed about block-specific modification functions as a first-class citizens.

dbolotin (changes requested):
# Context

Please read the following articles: https://www.anthropic.com/engineering/writing-tools-for-agents (especially "Optimizing tool responses for token efficiency" section)

At some point this will become a solution, however we are not yet there, and Claude Code is not the only platform we develop for:
https://www.anthropic.com/engineering/advanced-tool-use

**Solution:** please see my comment for the `get_block_stat` method .

# Help & domain concepts

Platforma has domain-specific concepts (PFrame, PColumn, axes, Result Pool) that agents won't know from pre-training. Without guidance, the agent guesses — and guesses wrong.

The MCP spec has three primitives: Resources (app-controlled), Prompts (user-controlled), Tools (model-controlled). Only tools let the agent autonomously fetch knowledge mid-task. Anthropic's own progressive disclosure pattern (search_tools with detail levels, deferred loading) confirms this approach.

Add a `get_help` tool — concise topic index, ~300 tokens per topic, referenced from tool descriptions:

```typescript
server.registerTool(
  "get_help",
  {
    description:
      "Get documentation about Platforma concepts (PFrame, PColumn, axes, " +
      "Result Pool, block state, etc). Call without topic to see the index.",
    inputSchema: {
      topic: z.string().optional().describe("Topic name. Omit for index."),
    },
    annotations: { readOnlyHint: true },
  },
  async ({ topic }) => {
    if (!topic) return textResult(TOPICS["index"]);
    return TOPICS[topic]
      ? textResult(TOPICS[topic])
      : errorResult(`Unknown topic. Available: ${Object.keys(TOPICS).join(", ")}`);
  },
);

// Other tool descriptions reference it without explaining concepts inline:
// "Returns PFrame/PTable handles. Use get_help('pframes') if unfamiliar."
```

Agent loads domain knowledge on demand — zero cost when it doesn't need it, ~300 tokens when it does.

lib/node/pl-mcp-server/src/tools/data-query.ts
```diff
@@ -0,0 +1,357 @@
...
+        maxColumns: z
+          .number()
+          .optional()
+          .default(30)
+          .describe("Max columns to return per PFrame/PTable summary (default 30)."),
+      },
+    },
+    async ({ projectId, blockId, maxColumns }) => {
+      const project = await ctx.getOpenedProject(projectId);
+      const state: any = await withTimeout(
```
dbolotin: PR should have no `any` as discussed with Alexander.

dbolotin (changes requested):
@AStaroverov Please do another round of review, a lot have changed since your last iteration.

blackcat:
> # Context
> Please read the following articles: https://www.anthropic.com/engineering/writing-tools-for-agents (especially "Optimizing tool responses for token efficiency" section)
> 
> ...
> 
> Add a `get_help` tool ...

Discussed. Skills should work better for our goals. Skill for blocks modified with MCP instructions. And skill for desktop added.

lib/node/pl-mcp-server/src/tools/types.ts
```diff
@@ -0,0 +1,156 @@
...
+  return runInNewContext(`(${expression})`, context, {
+    timeout,
+    filename: "transform",
+  });
+}
+
+const DEFAULT_STATE_TIMEOUT = 15_000;
+
+/** Race a promise against a timeout. Rejects with a descriptive error on timeout. */
+export function withTimeout<T>(
```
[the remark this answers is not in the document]
blackcat: done in 4be686302

lib/node/pl-mcp-server/src/tools/await.ts
```diff
@@ -0,0 +1,152 @@
...
+              });
+            }
+            return errorResult(
+              `Failed to get block state: ${e instanceof Error ? e.message : String(e)}`,
+            );
+          }
+
+          let data: unknown;
+          if (state.blockStorage) {
+            try {
```
vadimpiven: Please use https://github.com/milaboratory/platforma/blob/10568a826f7d3d55d3d9d402ba9df3516a0895bc/sdk/model/src/block_storage.ts#L175

blackcat: fixed in 8bcc649b4d2be5ce497b26ecc9076acd19513d44

lib/node/pl-mcp-server/src/tools/block-state.ts
```diff
@@ -0,0 +1,119 @@
...
+          .default(5000)
+          .describe("Timeout in ms for transform evaluation (default 5000)."),
+      },
+    },
+    async ({ projectId, blockId, transform, transformTimeout }) => {
+      const project = await ctx.getOpenedProject(projectId);
+      const state = await project.getBlockState(blockId).getValue();
+      let data: unknown = undefined;
+      if (state.blockStorage) {
+        try {
```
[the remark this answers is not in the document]
blackcat: 8bcc649b4d2be5ce497b26ecc9076acd19513d44

lib/node/pl-mcp-server/src/tools/block-state.ts
```diff
@@ -0,0 +1,119 @@
...
+        projectId: z.string().describe("Project ID"),
+        blockId: z.string().describe("Block ID"),
+        data: z.record(z.unknown()).describe("Block data object"),
+      },
+    },
+    async ({ projectId, blockId, data }) => {
+      const project = await ctx.getOpenedProject(projectId);
+      await project.mutateBlockStorage(
+        blockId,
+        { operation: "update-block-data", value: data },
```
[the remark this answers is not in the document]
blackcat: No I wasn't aware =(
This PR is already huge let's support them as separate feature.

lib/node/pl-mcp-server/src/tools/data-query.ts
```diff
@@ -0,0 +1,357 @@
...
+} {
+  const pFrames: { path: string; handle: string }[] = [];
+  const pTables: { path: string; handle: string }[] = [];
+  const hexHashPattern = /^[a-f0-9]{64}$/;
+
+  function walk(obj: unknown, path: string) {
+    if (obj === null || obj === undefined) return;
+    if (typeof obj === "string" && hexHashPattern.test(obj)) {
+      const lowerPath = path.toLowerCase();
+      if (lowerPath.includes("pframe")) {
```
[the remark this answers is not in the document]
blackcat: fixed in 3fe8442049911d69e2d84c8569ab26c8f28134bf

lib/node/pl-mcp-server/src/tools/data-query.ts
```diff
@@ -0,0 +1,357 @@
...
+
+  return { pFrames, pTables };
+}
+
+/**
+ * Converts PTableVector data to JSON-serializable arrays.
+ * Handles typed arrays (Int32Array, Float64Array, BigInt64Array, etc.)
+ * and marks NA/absent values.
+ */
+function vectorToJson(vector: PTableVector, rows: number): (string | number | null | "ABSENT")[] {
```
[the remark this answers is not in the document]
blackcat: fixed in  cddaca83db5b3b859a676d68403c5e838a8e8033

lib/node/pl-mcp-server/src/tools/data-query.ts
```diff
@@ -0,0 +1,357 @@
...
+      // For each unique PFrame, list columns (summary only)
+      const pFrameDriver = ctx.requireMl().internalDriverKit.pFrameDriver;
+      const pFrameSummaries = [];
+      for (const [handle, paths] of pFramesByHandle) {
+        try {
+          const columns = await pFrameDriver.listColumns(handle as PFrameHandle);
+          const mapped = columns.slice(0, MAX_COLUMNS).map((c) => ({
+            name: c.spec.name,
+            valueType: c.spec.valueType,
+            label: c.spec.annotations?.["pl7.app/label"],
```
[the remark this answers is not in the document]
blackcat: done in 6 places. 9a3bb63f6ca1ea34973f3e6d245d6e96e5439780

lib/node/pl-mcp-server/src/tools/data-query.ts
```diff
@@ -0,0 +1,357 @@
...
+      for (const [handle, paths] of pTablesByHandle) {
+        try {
+          const shape = await pFrameDriver.getShape(handle as PTableHandle);
+          const spec = await pFrameDriver.getSpec(handle as PTableHandle);
+          const mapped = spec.slice(0, MAX_COLUMNS).map((s: PTableColumnSpec, idx: number) => ({
+            index: idx,
+            type: s.type,
+            name: s.type === "column" ? s.spec.name : s.spec.name,
+            valueType: s.type === "column" ? s.spec.valueType : s.spec.type,
+            label: s.spec.annotations?.["pl7.app/label"],
```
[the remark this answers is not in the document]
blackcat: done in 9a3bb63f6ca1ea34973f3e6d245d6e96e5439780

lib/node/pl-mcp-server/src/tools/data-query.ts
```diff
@@ -0,0 +1,357 @@
...
+    },
+    async ({ pFrameHandle }) => {
+      const pFrameDriver = ctx.requireMl().internalDriverKit.pFrameDriver;
+      const columns = await pFrameDriver.listColumns(pFrameHandle as PFrameHandle);
+      return textResult(
+        columns.map((c) => ({
+          columnId: c.columnId,
+          name: c.spec.name,
+          valueType: c.spec.valueType,
+          label: c.spec.annotations?.["pl7.app/label"],
```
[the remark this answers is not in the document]
blackcat: done in 9a3bb63f6ca1ea34973f3e6d245d6e96e5439780

lib/node/pl-mcp-server/src/tools/data-query.ts
```diff
@@ -0,0 +1,357 @@
...
+          visibility: c.spec.annotations?.["pl7.app/table/visibility"],
+          axes: c.spec.axesSpec.map((a) => ({
+            name: a.name,
+            type: a.type,
+            label: a.annotations?.["pl7.app/label"],
```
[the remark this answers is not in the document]
blackcat: fix in 9a3bb63f6ca1ea34973f3e6d245d6e96e5439780

lib/node/pl-mcp-server/src/tools/data-query.ts
```diff
@@ -0,0 +1,357 @@
...
+        rows.push(columnVectors.map((col) => col[r]));
+      }
+
+      const columnHeaders = columnIndices.map((idx: number) => {
+        const s = spec[idx];
+        return {
+          index: idx,
+          type: s.type,
+          name: s.type === "column" ? s.spec.name : s.spec.name,
+          label: s.spec.annotations?.["pl7.app/label"],
```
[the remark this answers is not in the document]
blackcat: fix in 9a3bb63f6ca1ea34973f3e6d245d6e96e5439780

lib/node/pl-mcp-server/src/tools/logs.ts
```diff
@@ -0,0 +1,86 @@
...
+    async ({ projectId, blockId, lines, sampleId }) => {
+      const project = await ctx.getOpenedProject(projectId);
+      const state = await project.getBlockState(blockId).getValue();
+      if (!state.outputs)
+        return errorResult(
+          "Block has no outputs yet.",
+          "The block may not have been run. Use get_project_overview to check its calculationStatus, then run_block if needed.",
+        );
+
+      // Find log handles in outputs — look for the "logs" output
```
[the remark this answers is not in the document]
blackcat: fix in 9d1462f14537525d298a4b16d4ed4a88a9c820fd
Found lib/model/common/src/drivers/log.ts with some LogHandle helpers, so put mine there too. Not sure why Gleb placed it into the `drivers` folder.

lib/node/pl-mcp-server/src/tools/screenshot.ts
```diff
@@ -0,0 +1,48 @@
...
+        return errorResult(
+          "Screenshot capture is not available.",
+          "Make sure the MCP server is running inside Platforma Desktop and MCP connected properly. If everything is fine check Electron logs with get_app_log",
+        );
+      }
+      const base64Png = await ctx.callbacks.captureScreenshot();
+
+      if (savePath) {
+        const absPath = resolve(savePath);
+        await writeFile(absPath, Buffer.from(base64Png, "base64"));
```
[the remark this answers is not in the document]
blackcat: fixed in f2eb2db283eb15fb533b50549af31ac50a55aea3

lib/node/pl-mcp-server/src/tools/types.ts
```diff
@@ -0,0 +1,133 @@
...
+
+/**
+ * Estimate the number of LLM tokens needed to represent a value as JSON,
+ * without actually serializing it. Walks the object tree and sums up
+ * the JSON character lengths, then divides by ~4 chars/token.
+ *
+ * Returns token estimate, or a string like ">1234 (truncated at 10000 nodes)"
+ * if the object graph is too large to fully traverse.
+ */
+export function estimateTokens(value: unknown, nodeLimit = NODE_LIMIT): number | string {
```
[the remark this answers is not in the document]
blackcat: All for NODE_LIMIT. I'm not sure how huge the input could be. Gigabytes? We don't need them in AI agent context. Also traverse with limit will be faster.

lib/node/pl-mcp-server/src/tools/types.ts
```diff
@@ -0,0 +1,133 @@
...
+import type { MiddleLayer, Project, ProjectListEntry } from "@milaboratories/pl-middle-layer";
+import type { PlMcpServerCallbacks } from "../server";
+
+export interface AuthorMarker {
+  authorId: string;
+  localVersion: number;
+}
+
+export interface ToolContext {
+  getMl: () => MiddleLayer | undefined;
```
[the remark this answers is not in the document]
blackcat: done in e9cd6e0853c3024adc071d654e7b1a3524f270bb

lib/node/pl-mcp-server/src/server.ts
```diff
@@ -0,0 +1,265 @@
...
+  lastConnected?: string;
+}
+
+export interface PlMcpServerOptions {
+  /** MiddleLayer instance providing access to projects, blocks, etc. Optional — server can start without it. */
+  middleLayer?: MiddleLayer;
+  /** Port to listen on. */
+  port: number;
+  /** Secret path segment for URL security. */
+  secret: string;
```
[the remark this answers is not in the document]
blackcat: fixed in c8da14d68ef303eeb17c5e5ade012037fae131e3

tests/mcp-server/src/with-mcp.ts
```diff
@@ -0,0 +1,68 @@
...
+export interface McpTestContext {
+  client: Client;
+  ml: MiddleLayer;
+  serverUrl: string;
+}
+
+export async function withMcpServer(cb: (ctx: McpTestContext) => Promise<void>): Promise<void> {
+  const workFolder = path.resolve(import.meta.dirname, "..", "work", randomUUID());
+  const secret = randomUUID().replace(/-/g, "");
+  const port = await getFreePort();
```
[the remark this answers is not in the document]
blackcat: fixed in 3788570c0761bbe10b3b3989b9444e201fee1565

lib/node/pl-mcp-server/src/server.ts
```diff
@@ -0,0 +1,265 @@
...
+  private ml: MiddleLayer | undefined;
+  private port: number;
+  private readonly secret: string;
+  private readonly callbacks: PlMcpServerCallbacks;
+  private httpServer: Server | undefined;
+  private readonly transports = new Map<string, StreamableHTTPServerTransport>();
+
+  constructor(options: PlMcpServerOptions) {
+    this.ml = options.middleLayer;
+    this.port = options.port;
```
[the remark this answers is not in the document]
blackcat: fixed in 3788570c0761bbe10b3b3989b9444e201fee1565

lib/node/pl-mcp-server/src/tools/await.ts
```diff
@@ -0,0 +1,158 @@
...
+        if (block.calculationStatus === "Done") {
+          // Await stable block state with remaining time budget
+          const remaining = Math.max(deadline - Date.now(), 1000);
+          let state;
+          try {
+            state = await project
+              .getBlockState(blockId)
+              .awaitStableValue(AbortSignal.timeout(remaining));
+          } catch (e: unknown) {
+            if (e instanceof DOMException && e.name === "TimeoutError") {
```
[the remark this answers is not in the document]
blackcat: fixed in 972f731669628e99f18515c9d7021cd0c5455a8b

lib/node/pl-mcp-server/src/server.ts
```diff
@@ -0,0 +1,265 @@
...
+  }
+
+  async start(): Promise<void> {
+    if (this.httpServer) {
+      throw new Error("MCP server is already running");
+    }
+
+    const expectedPath = `/${this.secret}/mcp`;
+
+    this.httpServer = createServer(async (req: IncomingMessage, res: ServerResponse) => {
```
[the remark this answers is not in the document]
blackcat: simplified server start in a00cb35ea668851cd801cdb2dcc9ed6dfe4519ae

lib/model/common/src/drivers/log.ts
```diff
@@ -1,18 +1,30 @@
...
 /** Handle of the ready logs of a program. */
-export type ReadyLogHandle = `log+ready://log/${string}`;
+export type ReadyLogHandle = `${typeof READY_LOG_PREFIX}${string}`;
+
+/** Handle of logs. This handle should be passed
+ * to the driver for retrieving logs. */
+export type AnyLogHandle = LiveLogHandle | ReadyLogHandle;
+
+/** Type guard to check if a value is any kind of log handle. */
+export function isAnyLogHandle(value: unknown): value is AnyLogHandle {
```
vadimpiven, line 20: can we drop Any from names?

lib/node/pl-mcp-server/src/tools/data-query.ts
```diff
@@ -0,0 +1,327 @@
...
+ */
+async function resolveHandle(
+  handle: string,
+  driver: {
+    listColumns: (
+      h: PFrameHandle,
+    ) => Promise<
+      { spec: { name: string; valueType: string; annotations?: Record<string, string> } }[]
+    >;
+    getShape: (h: PTableHandle) => Promise<{ rows: number }>;
```
vadimpiven: Please do not call this, use listColumns for pframe and getSpec for ptable, getShape triggers join calculation which we want to aviod

lib/node/pl-mcp-server/src/tools/data-query.ts
```diff
@@ -0,0 +1,327 @@
...
+  }
+  return out;
+}
+
+/** Converts a PTableVector column to a JSON-serializable array using the SDK helper. */
+function vectorToJson(vector: PTableVector, rows: number): (string | number | null | "ABSENT")[] {
+  const result: (string | number | null | "ABSENT")[] = [];
+  for (let i = 0; i < rows; i++) {
+    const v = pTableValue(vector, i);
+    if (v === null) {
```
vadimpiven: do you need this if-else? why not just push v directly?

lib/node/pl-mcp-server/src/tools/data-query.ts
```diff
@@ -0,0 +1,327 @@
...
+          .describe("Timeout in ms for transform evaluation (default 5000)."),
+      },
+    },
+    async ({ pTableHandle, columns, offset, limit, maxLimit, transform, transformTimeout }) => {
+      const pFrameDriver = ctx.requireMl().internalDriverKit.pFrameDriver;
+      const handle = pTableHandle as PTableHandle;
+
+      let shape;
+      try {
+        shape = await pFrameDriver.getShape(handle);
```
vadimpiven: please call getShape after getSpec

lib/node/pl-mcp-server/src/server.ts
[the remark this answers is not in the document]
vadimpiven: I still vote for @modelcontextprotocol/sdk, I need a second opinion here, please ask someone else to review too

tests/mcp-server/src/with-mcp.ts
```diff
@@ -0,0 +1,49 @@
...
+      devBlockUpdateRecheckInterval: 300,
+      localSecret: MiddleLayer.generateLocalSecret(),
+      localProjections: [],
+      openFileDialogCallback: () => {
+        throw new Error("Not implemented.");
+      },
+    });
+    ml.addRuntimeCapability("requiresUIAPIVersion", 1);
+    ml.addRuntimeCapability("requiresUIAPIVersion", 2);
+    ml.addRuntimeCapability("requiresUIAPIVersion", 3);
```
vadimpiven, line 32: I think you also need requiresCreatePTable: 2

lib/model/common/src/drivers/log.ts
blackcat: There are more than 30 usages across SDK. And few in blocks. Better to change it in separate PR or not to change at all.

lib/model/common/src/drivers/log.ts
```diff
@@ -1,18 +1,30 @@
...
 /** Handle of the live logs of a program.
  * The resource that represents a log can be deleted,
  * in this case the handle should be refreshed. */
-export type LiveLogHandle = `log+live://log/${string}`;
+export type LiveLogHandle = `${typeof LIVE_LOG_PREFIX}${string}`;
 
 /** Handle of the ready logs of a program. */
-export type ReadyLogHandle = `log+ready://log/${string}`;
+export type ReadyLogHandle = `${typeof READY_LOG_PREFIX}${string}`;
+
```
AStaroverov: brand this types

lib/node/pl-mcp-server/src/tools/await.ts
```diff
@@ -0,0 +1,143 @@
...
+          .describe("Timeout in ms for transform evaluation (default 5000)."),
+      },
+    },
+    async ({ projectId, blockId, timeout, transform, transformTimeout }) => {
+      const project = await ctx.getOpenedProject(projectId);
+      const deadline = Date.now() + timeout;
+
+      while (Date.now() < deadline) {
+        const overview = await project.overview.getValue();
+        if (!overview) continue;
```
AStaroverov, line 43: are you sure that we don't add some delay?

lib/node/pl-mcp-server/src/tools/types.ts
```diff
@@ -0,0 +1,133 @@
+import { runInNewContext } from "node:vm";
+import type { MiddleLayer, Project, ProjectListEntry } from "@milaboratories/pl-middle-layer";
+import type { PlMcpServerCallbacks } from "../server";
+
+export interface AuthorMarker {
+  authorId: string;
```
AStaroverov: brand it

lib/node/pl-mcp-server/src/tools/types.ts
```diff
@@ -0,0 +1,133 @@
...
+
+/**
+ * Estimate the number of LLM tokens needed to represent a value as JSON,
+ * without actually serializing it. Walks the object tree and sums up
+ * the JSON character lengths, then divides by ~4 chars/token.
+ *
+ * Returns token estimate, or a string like ">1234 (truncated at 10000 nodes)"
+ * if the object graph is too large to fully traverse.
+ */
+export function estimateTokens(value: unknown, nodeLimit = NODE_LIMIT): number | string {
```
AStaroverov: pls start function from verb

AStaroverov: why types.ts include some logic?

AStaroverov: let's test functions with more or less complex logic

lib/node/pl-mcp-server/src/tools/types.ts
```diff
@@ -0,0 +1,133 @@
...
+/**
+ * Evaluate a JS expression in a sandboxed VM context.
+ * The expression has access to the provided variables.
+ */
+export function safeEval(
+  expression: string,
+  context: Record<string, unknown>,
+  timeout: number,
+): unknown {
+  return runInNewContext(`(${expression})`, context, {
```
AStaroverov: can you explain this concern? what do you want to achieve?

lib/node/pl-mcp-server/src/server.ts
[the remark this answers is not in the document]
blackcat: McpServer is the protocol layer — it registers tools, handles JSON-RPC messages, and manages the MCP protocol. It doesn't listen on a network port.

To actually receive HTTP requests, we need a transport layer. The SDK provides StreamableHTTPServerTransport which converts between HTTP request/response and the MCP protocol messages. But that transport still needs something to hand it the raw HTTP requests.

That "something" is the Node.js http.Server from createServer. It:

1. Binds to a port
2. Receives incoming HTTP connections
3. Routes them through our checks (secret path, origin validation)
4. Passes them to the transport's handleRequest

The chain is: http.Server → our handler → StreamableHTTPServerTransport.handleRequest() → McpServer

The SDK does offer createMcpExpressApp() from @modelcontextprotocol/express which wraps all of this (Express + transport + session management). But that adds Express as a dependency. Our custom createServer handler does the same thing with zero extra dependencies — just node:http.

Express peer-depends on express@^5.2.1 and @modelcontextprotocol/server@^2.0.0-alpha.2.

Two problems for us:
1. Express 5 — a heavy dependency we don't use anywhere else in the desktop app
2. @modelcontextprotocol/server@^2.0.0-alpha.2 — that's a v2 alpha. We're on @modelcontextprotocol/sdk@1.27.1. The SDK was split into separate packages (@modelcontextprotocol/server, @modelcontextprotocol/node, etc.) in v2, which is still in alpha

vadimpiven (approved):

lib/node/pl-mcp-server/src/server.ts
```diff
@@ -0,0 +1,271 @@
...
+  }>;
+  /** Disconnect from current server. */
+  disconnect?: () => Promise<void>;
+  /** Get detailed info about a specific block package. */
+  getBlockInfo?: (
+    registryUrl: string,
+    organization: string,
+    name: string,
+    version: string,
+  ) => Promise<unknown>;
```
AStaroverov, line 59: really unknown?

lib/node/pl-mcp-server/src/server.ts
```diff
@@ -0,0 +1,271 @@
...
+  ) => Promise<{ status: string; message: string }>;
+  /** Get current connection status. */
+  getConnectionStatus?: () => Promise<{
+    connected: boolean;
+    type?: string;
+    addr?: string;
+    login?: string;
+  }>;
+  /** Disconnect from current server. */
+  disconnect?: () => Promise<void>;
```
AStaroverov, line 52: you mean `onDisconnect` ?

lib/node/pl-mcp-server/src/server.ts
```diff
@@ -0,0 +1,271 @@
...
+
+  constructor(options: PlMcpServerOptions) {
+    this.ml = options.middleLayer ?? null;
+    this.port = options.port;
+    this.secret = options.secret;
+    this.callbacks = options.callbacks ?? {};
+  }
+
+  /** Set or update the MiddleLayer instance (e.g. after connecting to a server). */
+  setMiddleLayer(ml: MiddleLayer | null | undefined) {
```
AStaroverov, line 99: do we really need this method? mb better have readonly ml?

lib/node/pl-mcp-server/src/server.ts
```diff
@@ -0,0 +1,271 @@
...
+    this.ml = ml ?? null;
+  }
+
+  get url(): string {
```
AStaroverov, line 103: getUrl

lib/node/pl-mcp-server/src/tools/types.ts
blackcat: Estimate is a verb

lib/model/common/src/drivers/log.ts
blackcat: fixed in 657affecd467f19b3d684e39c295db752ae381bf

lib/node/pl-mcp-server/src/tools/await.ts
blackcat: fixed in 657affecd467f19b3d684e39c295db752ae381bf

lib/node/pl-mcp-server/src/tools/types.ts
blackcat: fixed in 657affecd467f19b3d684e39c295db752ae381bf

lib/node/pl-mcp-server/src/tools/types.ts
blackcat: fixed in 657affecd467f19b3d684e39c295db752ae381bf

blackcat: fixed in 657affecd467f19b3d684e39c295db752ae381bf

lib/node/pl-mcp-server/src/tools/types.ts
blackcat: Forgot to move this to the QuickJS VM. Thank you! 
Fixed in ef258fabd4f77a41306e38716f857532d621237c

blackcat: Block data could be huge, no **HUGE**. We don't want to load full json into the context. Thats why we have `transform` argument. It allows claude inject JS to query exact data.

This https://github.com/milaboratory/platforma/blob/feature/mcp-server/lib/node/pl-mcp-server/src/tools/block-state.ts#L43 will help you.

AStaroverov (approved):

dbolotin (approved):

PR #1553, opened 2026-04-08
MILAB-5815: resource signature support

rfiskov:

This PR wires up resource signature support across the `pl-client` stack: new `resourceSignature`, `valueSignature`, and `errorSignature` fields flow from the proto layer through `type_conversion.ts` into `types.ts`; `PlTransaction` stores and attaches signatures to every resource/field operation; and a new `SignatureCache` persists signatures across transactions, seeded at init from `ListUserResources`. A `setDefaultColor` path is also added so write transactions carry the correct access color.

- The \"check is abort error (active)\" test always creates a resource with all-zero data due to `Math.random() & 0xff` (bitwise AND truncates the float to 0 before the AND); the correct pattern `(Math.random() * 256) & 0xff` is already used in the sibling test.
</details>

Safe to merge after fixing the test byte-generation bug; all other findings are style/P2.

One P1 finding in the test file: `Math.random() & 0xff` always yields 0, meaning the abort-error test never exercises random payloads as intended. The production implementation (transaction, client, cache) is sound. Remaining findings (unbounded cache, redundant await, undocumented clear() responsibility) are all P2.

lib/node/pl-client/src/core/ll_transaction.test.ts — byte-generation bug on line 190.
</details>

<details open><summary><h3>Vulnerabilities</h3></summary>

No security concerns identified. Signatures are opaque byte arrays passed through from the server without local manipulation, and the cache is not shared across clients.
</details>

| Filename | Overview |
|----------|----------|
| lib/node/pl-client/src/core/ll_transaction.test.ts | Tests updated for color proof / setDefaultColor flow; contains `Math.random() & 0xff` (always 0) in the abort test and redundant double-await on already-resolved responses. |
| lib/node/pl-client/src/core/signature_cache.ts | New cross-transaction signature cache backed by a plain Map; works correctly but has no eviction policy, which may grow unbounded in long-lived clients. |
| lib/node/pl-client/src/core/client.ts | Adds SignatureCache instantiation, root signature seeding from ListUserResources, and default color proof injection; documented clear()-on-auth-error contract is not wired up. |
| lib/node/pl-client/src/core/transaction.ts | PlTransaction extended with signature storage, retrieval, and propagation for all resource/field operations; implementation looks correct. |
| lib/node/pl-client/src/core/types.ts | Adds optional `resourceSignature` to BasicResourceData and `valueSignature`/`errorSignature` to FieldData; straightforward and safe additions. |
| lib/node/pl-client/src/core/type_conversion.ts | Maps new proto signature fields into domain types in protoToResource and protoToField; correct and complete. |
| lib/node/pl-client/src/core/errors.ts | Adds `isUnimplementedError` helper used to gracefully fall back from ListUserResources; simple and correct. |
| lib/node/pl-client/src/index.ts | Re-exports new SignatureCache from the package index; straightforward. |
| lib/node/pl-client/src/test/test_config.ts | Minor test infrastructure updates; no issues found. |
| lib/node/pl-drivers/proto/shared/protodep.toml | Proto dependency config; minimal change, safe. |
| .changeset/cuddly-dancers-greet.md | Comprehensive major-version changeset documenting all proto field additions, renames, and new APIs; well-written. |

</details>

</details>

<details open><summary><h3>Comments Outside Diff (1)</h3></summary>

1. `lib/node/pl-client/src/core/ll_transaction.test.ts`, line 190-199 ([link](https://github.com/milaboratory/platforma/blob/c05891e68c9b7abb2cf422867b13695834bdf94c/lib/node/pl-client/src/core/ll_transaction.test.ts#L190-L199)) 

   <a href="#">P1</a> **`Math.random() & 0xff` always produces 0**

   `Math.random()` returns a float in `[0, 1)`. JavaScript's bitwise `&` first converts the operand to a 32-bit integer via `ToInt32`, which truncates the fractional part — so `Math.random() & 0xff` is always `0`. The "check is abort error (active)" test always creates a resource with an all-zeros payload. The corresponding "check timeout error type (active)" test on line 118 uses `(Math.random() * 256) & 0xff` and is correct.

   

   `````markdown
   Path: lib/node/pl-client/src/core/ll_transaction.test.ts
   Line: 190-199

   Comment:
   **`Math.random() & 0xff` always produces 0**

   `Math.random()` returns a float in `[0, 1)`. JavaScript's bitwise `&` first converts the operand to a 32-bit integer via `ToInt32`, which truncates the fractional part — so `Math.random() & 0xff` is always `0`. The "check is abort error (active)" test always creates a resource with an all-zeros payload. The corresponding "check timeout error type (active)" test on line 118 uses `(Math.random() * 256) & 0xff` and is correct.

   

   `````
   </details>

</details>

`````markdown
Path: lib/node/pl-client/src/core/ll_transaction.test.ts
Line: 190-199

Comment:
**`Math.random() & 0xff` always produces 0**

`Math.random()` returns a float in `[0, 1)`. JavaScript's bitwise `&` first converts the operand to a 32-bit integer via `ToInt32`, which truncates the fractional part — so `Math.random() & 0xff` is always `0`. The "check is abort error (active)" test always creates a resource with an all-zeros payload. The corresponding "check timeout error type (active)" test on line 118 uses `(Math.random() * 256) & 0xff` and is correct.

```suggestion
    const rData = Uint8Array.from([
      (Math.random() * 256) & 0xff,
      (Math.random() * 256) & 0xff,
      (Math.random() * 256) & 0xff,
      (Math.random() * 256) & 0xff,
      (Math.random() * 256) & 0xff,
      (Math.random() * 256) & 0xff,
      (Math.random() * 256) & 0xff,
      (Math.random() * 256) & 0xff,
    ]);
```

---

Path: lib/node/pl-client/src/core/ll_transaction.test.ts
Line: 141

Comment:
**Redundant `await` on already-resolved value**

`createResponse` is obtained via `const createResponse = await tx.send(...)` (line 129), so it is already the resolved `ServerMessageResponse` value, not a `Promise`. The inner `await createResponse` on this line re-awaits a non-Promise, which works but is misleading. The same pattern appears in the "check is abort error (active)" test at line 213.

```suggestion
   const createResp = createResponse.resourceCreateValue;
```

---

Path: lib/node/pl-client/src/core/signature_cache.ts
Line: 3-4

Comment:
**Unbounded cache with no eviction policy**

`SignatureCache` uses a plain `Map` keyed by resource ID with no size cap or TTL. In a long-running process where many resources are created or observed, the store will grow without bound. Consider capping it with an `LRUCache` (already a dependency via `lru-cache`) or adding a `maxSize` option, consistent with the existing `resourceDataCache` in `PlClient`.

---

Path: lib/node/pl-client/src/core/client.ts
Line: 216-219

Comment:
**`clear()` on auth errors is documented but not implemented**

The JSDoc says _"Call clear() on auth errors to invalidate stale signatures"_, but there is no call site in `PlClient` (or the auth refresh path) that actually invokes `this._signatureCache.clear()`. Callers who rely on this cache through the public `signatureCache` getter would need to know to call it themselves. Either wire it up automatically in the auth-error handling path, or make the documentation more explicit that this is entirely the caller's responsibility.

`````

</details>

<sub>Reviews (1): Last reviewed commit: ["MILAB-5815: fix: test sign propogation"](https://github.com/milaboratory/platforma/commit/c05891e68c9b7abb2cf422867b13695834bdf94c) | [Re-trigger Greptile](https://app.greptile.com/api/retrigger?id=27717092)</sub>

### CODE REVIEW

File: lib/node/pl-client/src/core/signature_cache.ts [hunk 1]
```diff
@@ -0,0 +1,44 @@
+import { LRUCache } from "lru-cache";
+import type { ResourceId, ResourceSignature, SignedResourceId } from "./types";
+
+const DEFAULT_MAX_ENTRIES = 100_000;
+
+/** Cross-transaction cache for resource signatures.
+ * Keyed by ResourceId, stores opaque ResourceSignature bytes.
+ * Uses LRU eviction to prevent unbounded memory growth in long-running clients. */
+export class SignatureCache {
+  private readonly store: LRUCache<ResourceId, ResourceSignature>;
```
dbolotin: `LRUCache` is not a proper structure here and creates a fragile time bomb. All the users of the pl-client rely on this structure to preserve signatures for the resource id's that they store elsewhere, given the nature of the cache some signatures will be eventually evicted (and resource turnaround in real projects is pretty high) creating irrecoverable state for the whole system. We have a `global resource id` type and associated logic with explicit call in the transaction where user resolves the global resource IDs, expressing the intent to get something from the system that can be stored "persistently" to be later used for resource retrieval. We must rely on this mechanism and explicitly store signatures in those places where resource ids are persisted locally.

dbolotin (changes requested):

lib/node/pl-client/src/core/errors.ts
```diff
@@ -125,6 +137,13 @@ export class UnauthenticatedError extends Error {
   }
 }
 
+export class PermissionDeniedError extends Error {
+  name = "PermissionDeniedError";
+  constructor(message: string) {
+    super("PermissionDenied: " + message);
+  }
+}
+
```
DenKoren: Looks like this class is not compatible with isPermissionDenied check:
```typescript
isPermissionDenied(new PermissionDeniedError("aaaa")) // returns false?
```

lib/node/pl-client/src/core/transaction.ts
```diff
@@ -276,6 +303,60 @@ export class PlTransaction {
...
+    if ("fields" in result && result.fields) {
+      for (const f of result.fields) {
+        if (!isNullResourceId(f.value)) this.storeSignature(f.value, f.valueSignature);
+        if (!isNullResourceId(f.error)) this.storeSignature(f.error, f.errorSignature);
+      }
+    }
+  }
+
+  /** Set default color proof for subsequent resource creation requests */
+  public setDefaultColor(colorProof: ColorProof): void {
```
DenKoren: nit: what do you think of accepting not only plain color, but also a global resource ID here? So transaction automatically takes signature out of it and use it as color proof?
Then it would be easy to instantiate a transaction without additional calls that extract signature from resource.

PR #1573, opened 2026-04-20
Full migratione from cjs to es

AStaroverov:

This PR completes the migration of 97 files across the monorepo from CommonJS to ESM, changing the default `createRolldownNodeConfig` output from `["es", "cjs"]` to `["es"]`, adding `"type": "module"` to package manifests, polyfilling `__dirname` in CLI entry points, and replacing the hand-written dual-format `resolve-helper` files with a single TypeScript source that implements a two-stage CJS→ESM resolver. The most notable functional change is the replacement of Zod-based `GlobalOverviewReg` schema validation with a manual migration function that skips inner-field validation.
</details>

Safe to merge; all findings are P2 style/robustness concerns that do not affect current functionality.

All three inline comments are P2: dead code in the ESM resolver, a loss of deep runtime validation (intentional trade-off for removing Zod from the parse path), and a future fragility in `require.resolve` for model packages that don't yet have exports fields. No regressions in the primary migration path are evident.

lib/node/resolve-helper/src/index.ts (dead code), tools/block-tools/src/v2/registry/schema_public.ts (validation regression)
</details>

| Filename | Overview |
|----------|----------|
| lib/node/resolve-helper/src/index.ts | New TypeScript ESM resolver: replaces hand-written CJS/ESM dual files with a full implementation; dead code on line 181 (`"default"` condition checked twice). |
| tools/block-tools/src/v2/registry/schema_public.ts | Replaces Zod-based GlobalOverviewReg schemas with plain TypeScript types + manual migration; loses runtime validation of `id` and description fields. |
| tools/block-tools/src/cmd/build-model.ts | Migrates from synchronous `require()` to `createRequire().resolve()` + dynamic `import()` for ESM support; fragile for future pure-ESM model packages with exports field. |
| tools/ts-builder/src/configs/utils/createRolldownNodeConfig.ts | Default output formats changed from `["es", "cjs"]` to `["es"]` — the central ESM-only change for the build infrastructure. |
| tools/block-tools/bin/run.js | Added standard `__filename`/`__dirname` ESM polyfill needed by `oclif.execute({ dir: __dirname })`. |
| lib/node/resolve-helper/package.json | Moved from hand-written `src/index.js`/`.mjs` dual output to a proper `ts-builder`-built ESM package; added standard build scripts and engine requirement `node >= 22`. |
| tools/block-tools/package.json | Migrated from Vite dual CJS/ESM build to `ts-builder` ESM-only build; updated oclif target path from `./dist/cli.js` to `./dist/cmd/index.js`. |

</details>

</details>

`````markdown
Path: lib/node/resolve-helper/src/index.ts
Line: 181

Comment:
**Dead code — `"default"` is already handled by the `CONDITIONS` loop**

`CONDITIONS` is `["node", "import", "default"]`, so when the loop at lines 175–180 runs, it already tries `"default"` and — if found — calls `resolveConditions(obj.default, subpath)`. If that call returns `undefined`, the loop finishes and falls through; the `if ("default" in obj)` guard on line 181 is then entered, but the call returns exactly the same `undefined` value. This line can never alter the final result and is unreachable in any meaningful way.

```suggestion
  return undefined;
```

---

Path: tools/block-tools/src/v2/registry/schema_public.ts
Line: 98-119

Comment:
**Loss of runtime validation for required fields**

The previous Zod-based `GlobalOverviewReg.parse()` validated every field including `id`, description objects (`BlockPackDescriptionManifest`), and SHA256 hashes, throwing a structured error on any mismatch. The new `migrateGlobalOverviewEntry` just casts these with `as`:

- `id: o.id as BlockPackIdNoVersion` — if `o.id` is `undefined` or malformed, the cast succeeds but downstream accesses on `id.organization` / `id.name` will throw cryptic `TypeError`s rather than a clear parse error.
- `description` values inside `latestByChannel` entries are not checked against `BlockPackDescriptionManifest` at all.

Consider adding at least a guard for the `id` field, e.g.:

```typescript
if (!o.id || typeof o.id !== "object")
  throw new Error(`Invalid global overview entry: missing or invalid id`);
```

---

Path: tools/block-tools/src/cmd/build-model.ts
Line: 48-51

Comment:
**`require.resolve` may fail for pure-ESM model packages with an `exports` field**

`createRequire(import.meta.url).resolve(modulePath)` uses the CJS resolution algorithm, which only honours `require` / `node` export conditions. If a model package under `flags.modulePath` exposes only an `import` condition (e.g. `"exports": { ".": { "import": "./dist/index.js" } }`), `require.resolve` will throw `ERR_PACKAGE_PATH_NOT_EXPORTED` and the `import()` call is never reached.

The current model packages (e.g. `enter-numbers/model`) rely solely on `"main"` (no `exports` field), so this works today. But as packages migrate further, this becomes fragile. Using `import.meta.resolve` (Node ≥ 20.6) or the workspace's own `tryResolve` helper from `@milaboratories/resolve-helper` would handle both CJS and ESM exports correctly.

```typescript
// Safer alternative (Node ≥ 20.6 or polyfilled via resolve-helper):
const resolved = await import.meta.resolve(pathToFileURL(modulePath).href);
let { model, platforma } = await import(resolved);
```

`````

</details>

<sub>Reviews (1): Last reviewed commit: ["full migration from cjs to es"](https://github.com/milaboratory/platforma/commit/4a09cbd75830e35f66c10f8bd0f57b89a785ad81) | [Re-trigger Greptile](https://app.greptile.com/api/retrigger?id=28964475)</sub>

### CODE REVIEW

File: etc/blocks/blob-url-custom-protocol/block/index.js [hunk 1]
```diff
@@ -1,15 +1,14 @@
-const blockTools = require('@platforma-sdk/block-tools');
+import { fileURLToPath } from 'node:url';
+import { dirname } from 'node:path';
+import * as blockTools from '@platforma-sdk/block-tools';
 
-async function loadBlockDescription() {
+const __dirname = dirname(fileURLToPath(import.meta.url));
```
kevindetry-milaboratories, line 5: `import.meta.dirname` couldn't be used here?

PR #1574, opened 2026-04-21
MILAB-5815: resource signatures support in ML

DenKoren:
- Signed resource is a string encoding resource ID and signature.
- types.ts file in pl-client now has 'types split': world of bigints (LocalResourceId, GlobalResource, AnyResourceId and so on) and word of signed resource ID which has its own supportive Optional type and Null type/const
- everything that earlier was indexed by GlobalResourceId (bigint with ID got from server) is now indexed by SignedResourceId (explicitly, with type update in the code). All contracts that accept resource ID from outer code (pl-client first of all) takes SignedResourceId explicitly. LRU caches in pl-client work as before, but since they now keep data indexed by signed IDs, we will not have any problems with signatures mixing between trees.
- ML now isolates upper level code from resource ID, providing clear abstraction of ProjectId, which is derived from ResourceId, but is safe for persisting and reusable across client sessions. All methods in ML now expect ProjectId and return ProjectId.
- To implement this, ML got its own cache of projects, that allows to transform ProjectId into SignedResourceId and make a correct calls to pl-client transaction methods.
- Since loosing a signature is now breaking action, attempt to construct SignedResourceId via 'as SignedResourceId' type hardening is forbidden: I extended ts-builder with new feature that analyzes code and prevents devs using this across code. Main rule: If you don't know where to take signed ID or signature for your ID - find a correct way to take it instead of passing <ID>|<empty signature>  to the code hoping it will work.
  - This strictness allowed to catch several hidden bugs that might have strike later during real usage of Desktop App.
- We still 'leak' project ID to the upper level, and it is TODO work to make two worlds (above ML and below one) completely decoupled in terms of type. For now coding still requires discipline to not encode signatures into persisted data (i.e. file names on disk).
  - I.e. PFrames use resource IDs for blobID generation.
  - PlTreeEntry still provides public access to RID field. Hiding it to private causes too much pain now.
  - All the rest (to best of my knowledge) is decoupled and hidden.

This PR threads resource signatures throughout the pl-client/pl-drivers/pl-middle-layer stack. It introduces a new `SignedResourceId` string type (`"NG:0x…|<sigHex>"`), propagates signatures from all proto responses into resource IDs, auto-sets the default color proof from the client root's signature on writable transactions, and adds `PermissionDeniedError` for the new server-side enforcement surface.

- **P1 (`transaction.ts` line 1011):** `getKValueJsonIfExists` calls `getKValueString` instead of `getKValueStringIfExists` — the function never returns `undefined` and always throws on a missing key, defeating its `IfExists` contract.
</details>

Safe to merge after fixing the `getKValueJsonIfExists` bug; the signature-threading logic is correct.

One P1 defect: `getKValueJsonIfExists` calls the wrong helper and silently breaks any caller that relies on `undefined` being returned for a missing key. All other findings are P2. The core signature-propagation logic, new error class, and driver updates look correct.

`lib/node/pl-client/src/core/transaction.ts` (P1 bug in `getKValueJsonIfExists`); `lib/node/pl-drivers/src/drivers/helpers/ls_storage_entry.ts` (intentional but undocumented unsigned-ID round-trip).
</details>

| Filename | Overview |
|----------|----------|
| lib/node/pl-client/src/core/types.ts | Introduces `SignedResourceId` (string format `"NG:0x…|<sigHex>"`) and `ResourceSignature` branded types, along with `bigintToResourceId`, `parseSignedResourceId`, `toResourceSignature`, and base64url helpers — the core of the signature feature. |
| lib/node/pl-client/src/core/transaction.ts | Adds `toSignedResourceId`/`toSignedFieldId`/`toSignedErrorRef` helpers and threads signatures through all protocol messages; contains a P1 bug in `getKValueJsonIfExists` that calls `getKValueString` instead of `getKValueStringIfExists`, breaking the IfExists contract. |
| lib/node/pl-client/src/core/client.ts | Auto-sets the default color proof from the client root's signature at transaction open time (writable txs only), and threads the root's signature through `bigintToResourceId`. |
| lib/node/pl-client/src/core/errors.ts | Adds `PermissionDeniedError` class and `isPermissionDenied` detection function (mirroring existing `UnauthenticatedError` pattern), and wires both into `rethrowMeaningfulError`. |
| lib/node/pl-client/src/core/type_conversion.ts | Refactors `protoToResource` and `protoToField` to extract signatures from proto messages and embed them in `SignedResourceId` values; adds `protoIdToOptionalResourceId` helper. |
| lib/node/pl-drivers/src/drivers/helpers/download_remote_handle.ts | Embeds resource signature as a base64url segment in blob+remote URLs and parses it back on download; signer still covers the full content including the signature segment. |
| lib/node/pl-drivers/src/drivers/helpers/ls_storage_entry.ts | Updates `createRemoteStorageHandle` to strip the signature (only globalId stored in handle), and `parseRemoteStorageHandle` reconstructs an unsigned `ResourceId`; intentional but may cause permission errors if strict-mode signatures are required for storage resource ops. |
| lib/node/pl-drivers/src/drivers/download_blob/blob_key.ts | Updates `blobKey` to use `parseSignedResourceId` rather than the broken `BigInt(rId)` cast, and `pathToBlobInfo` to produce proper `SignedResourceId` from numeric filename. |
| lib/node/pl-client/src/core/ll_transaction.test.ts | Adds `getRootSignature` helper and wires `setDefaultColor` into the active-timeout and abort tests; fallback to empty `Uint8Array` on missing signature is a silent no-op or will cause confusing failures. |
| lib/node/pl-tree/src/test_utils.ts | Updates test fixture resource IDs to use `bigintToResourceId(...)` to produce properly formatted `SignedResourceId` strings. |
| lib/node/pl-middle-layer/src/pool/data.ts | Imports `anyResourceIdToBigint` for extracting raw bigint from the new string `ResourceId` type when computing hashes. |
| lib/node/pl-drivers/src/drivers/download_blob_url/driver.ts | Switches to `parseSignedResourceId` for extracting global ID when building file paths, replacing the broken direct `BigInt(id)` cast. |
| lib/node/pl-drivers/src/drivers/download_blob_url/driver_id.ts | Updated to use `parseSignedResourceId` for keying download tasks, correctly ignoring the signature in the map key. |

</details>

</details>

<details open><summary><h3>Comments Outside Diff (2)</h3></summary>

1. `lib/node/pl-client/src/core/transaction.ts`, line 1009-1015 ([link](https://github.com/milaboratory/platforma/blob/2bfef9d795eb992e26376c346e72bfd915fe2dfa/lib/node/pl-client/src/core/transaction.ts#L1009-L1015)) 

   <a href="#">P1</a> **`getKValueJsonIfExists` always throws on missing key**

   `getKValueString` (line 1011) returns `Promise<string>` and throws a not-found error when the key is absent — it never returns `undefined`. The `if (str === undefined)` guard is therefore dead code and the function will never return `undefined` as its signature promises, completely defeating the `IfExists` semantics.

   The fix is to call `getKValueStringIfExists` instead of `getKValueString`.

   `````markdown
   Path: lib/node/pl-client/src/core/transaction.ts
   Line: 1009-1015

   Comment:
   **`getKValueJsonIfExists` always throws on missing key**

   `getKValueString` (line 1011) returns `Promise<string>` and throws a not-found error when the key is absent — it never returns `undefined`. The `if (str === undefined)` guard is therefore dead code and the function will never return `undefined` as its signature promises, completely defeating the `IfExists` semantics.

   The fix is to call `getKValueStringIfExists` instead of `getKValueString`.

   `````
   </details>

2. `lib/node/pl-drivers/src/drivers/helpers/ls_storage_entry.ts`, line 84 ([link](https://github.com/milaboratory/platforma/blob/2bfef9d795eb992e26376c346e72bfd915fe2dfa/lib/node/pl-drivers/src/drivers/helpers/ls_storage_entry.ts#L84)) 

   <a href="#">P2</a> **Remote storage handle reconstructed without signature**

   `parseRemoteStorageHandle` rebuilds the resource ID as `bigintToResourceId(BigInt(resourceId))` — no signature — because the `remote://name/id` handle format (intentionally) stores only the numeric global ID. If the server enforces signature checking for storage-resource operations, any transaction using this ID will receive a `PERMISSION_DENIED` error. Consider documenting that unsigned IDs are intentional here, or plumbing the signature into the handle format (similar to how `download_remote_handle.ts` embeds `resourceSig` in the URL).

   `````markdown
   Path: lib/node/pl-drivers/src/drivers/helpers/ls_storage_entry.ts
   Line: 84

   Comment:
   **Remote storage handle reconstructed without signature**

   `parseRemoteStorageHandle` rebuilds the resource ID as `bigintToResourceId(BigInt(resourceId))` — no signature — because the `remote://name/id` handle format (intentionally) stores only the numeric global ID. If the server enforces signature checking for storage-resource operations, any transaction using this ID will receive a `PERMISSION_DENIED` error. Consider documenting that unsigned IDs are intentional here, or plumbing the signature into the handle format (similar to how `download_remote_handle.ts` embeds `resourceSig` in the URL).

   `````
   </details>

</details>

`````markdown
Path: lib/node/pl-client/src/core/transaction.ts
Line: 1009-1015

Comment:
**`getKValueJsonIfExists` always throws on missing key**

`getKValueString` (line 1011) returns `Promise<string>` and throws a not-found error when the key is absent — it never returns `undefined`. The `if (str === undefined)` guard is therefore dead code and the function will never return `undefined` as its signature promises, completely defeating the `IfExists` semantics.

The fix is to call `getKValueStringIfExists` instead of `getKValueString`.

---

Path: lib/node/pl-drivers/src/drivers/helpers/ls_storage_entry.ts
Line: 84

Comment:
**Remote storage handle reconstructed without signature**

`parseRemoteStorageHandle` rebuilds the resource ID as `bigintToResourceId(BigInt(resourceId))` — no signature — because the `remote://name/id` handle format (intentionally) stores only the numeric global ID. If the server enforces signature checking for storage-resource operations, any transaction using this ID will receive a `PERMISSION_DENIED` error. Consider documenting that unsigned IDs are intentional here, or plumbing the signature into the handle format (similar to how `download_remote_handle.ts` embeds `resourceSig` in the URL).

---

Path: lib/node/pl-client/src/core/ll_transaction.test.ts
Line: 11-19

Comment:
**Silent fallback to empty color proof in test helper**

`getRootSignature` returns `new Uint8Array(0)` when no root with a `resourceSignature` entry is found (e.g. against an older server or in a non-strict mode environment). That empty array is then passed directly as `colorProof` in `setDefaultColor`. An empty color proof is either silently ignored or rejected by the server; in either case the test will pass for the wrong reason or fail with a confusing error. Consider throwing or skipping the test when no signature is available.

`````

</details>

<sub>Reviews (1): Last reviewed commit: ["MILAB-5815: resource signatures support ..."](https://github.com/milaboratory/platforma/commit/2bfef9d795eb992e26376c346e72bfd915fe2dfa) | [Re-trigger Greptile](https://app.greptile.com/api/retrigger?id=29097387)</sub>

### CODE REVIEW

File: lib/model/middle-layer/src/project_list.ts [hunk 1]
```diff
@@ -2,9 +2,7 @@ import type { ProjectMeta } from "./project";
 
 /** Represents single entry in the list of projects owned by current user */
 export interface ProjectListEntry {
-  /** Project resource ID. */
-  rid: bigint;
-  /** Internal (user-specific) project id */
+  /** Unique project identifier in middle layer. Use to operate with given project. */
   id: string;
```
vadimpiven: can we make this a branded type?

lib/node/pl-client/src/core/transaction.ts
```diff
@@ -276,6 +302,40 @@ export class PlTransaction {
...
+    const signed = toResourceIdAndSignature(base.resourceId);
+    return {
+      resourceId: signed.resourceId,
+      fieldName: base.fieldName,
+      resourceSignature: signed.resourceSignature,
+    };
+  }
+
+  private toSignedErrorRef(rId: AnyResourceRef): {
+    errorResourceId: bigint;
```
vadimpiven, line 332: maybe we need branded type for resourceId?

lib/node/pl-client/src/core/type_conversion.ts
```diff
@@ -26,12 +25,20 @@ function resourceIsDeleted(proto: Resource): boolean {
   return proto.deletedTime !== undefined && proto.deletedTime.seconds !== 0n;
 }
 
+function protoIdToOptionalResourceId(id: bigint, signature?: Uint8Array): OptionalSignedResourceId {
+  if (id === 0n) return NullSignedResourceId;
```
vadimpiven: maybe we need NullResourceId const?

lib/node/pl-client/src/core/types.ts
```diff
@@ -234,16 +243,118 @@ export function resourceIdFromString(str: string): OptionalAnyResourceId | undef
...
+// more details here: https://egghead.io/blog/using-branded-types-in-typescript
+/** Opaque authorization signature attached to a resource. */
+declare const __resource_signature_type__: unique symbol;
+export type ResourceSignature = Uint8Array & { readonly [__resource_signature_type__]: true };
+
+/**
+ * Signed resource id is "<global ID>|<resource signature hex>", encoded as string
+ * (e.g. "NG:0x123EC|1234567890abcdef")
+ */
+export type SignedResourceId = string & { __resource_id_type__: "signed" };
```
vadimpiven: @AStaroverov should existing brand type be used here maybe, what do you think?

lib/model/middle-layer/src/project_list.ts
DenKoren: Took ProjectId into @milaboratories/pl-model-common. Shared this branded type between node and model levels.
Thanks for making me do this in the end :D

lib/node/pl-client/src/core/user_resources.ts
```diff
@@ -0,0 +1,338 @@
...
+    }
+
+    const result = new Map<string, StorageInfo>();
+    for (const entry of [...v1Entries, ...v2Entries]) {
+      result.set(entry.storageId, entry);
+    }
+    return result;
+  }
+
+  // --- Legacy path: named resources ---
```
vadimpiven: please drop dividers, AI generates new style every time

lib/node/pl-client/src/core/transaction.ts
DenKoren: This is low-level helpers for conversion from/to gRPC plain types. No brading here, as it kills the idea of this function.

Branding is the thing exposed to outside in public interfaces. We have branded types for various resource ID shapes (plain bigint with local/global branding + signed ID with its own branding)

lib/node/pl-client/src/core/type_conversion.ts
DenKoren: We already have it. Fixed.

lib/node/pl-client/src/core/types.ts
DenKoren: Just found what you mean. Modified types.ts to use new helper. Thanks.

lib/node/pl-middle-layer/src/middle_layer/middle_layer.ts
```diff
@@ -202,64 +248,53 @@ export class MiddleLayer {
...
-      tx.createField(field(this.projectListResourceId, id), "Dynamic", newPrj);
+      // Attach to project list with a random UUID field name
+      tx.createField(field(this.projectListResourceId, randomUUID()), "Dynamic", newPrj);
       await tx.commit();
-      return await toGlobalResourceId(newPrj);
     });
     await this.projectListTree.refreshState();
-    return resource;
+
+    const signedRid = await newPrj!.globalId;
```
vadimpiven: please never use ! in code, use proper ifs

tools/ts-builder/src/commands/linter.ts
```diff
@@ -12,6 +12,102 @@ function getDefaultConfigPath(): string {
   return join(__dirname, "..", "configs", "oxlint-base.json");
 }
 
+/** Patterns that ban 'as T' type assertions for specific branded types. */
+const BANNED_TYPE_ASSERTIONS: { pattern: RegExp; message: string }[] = [
```
vadimpiven, line 16: I am not sure, but maybe better make this a eslint extension and add to default oxlint config?

vadimpiven (approved):

tools/ts-builder/src/commands/linter.ts
DenKoren: oxlint allows only banning `as Type` conversions globally, for all the code and types. This is too strict for our codebase and causes crazy amount of refactoring required.

If you know better way - could you please tell me how? I did not find one (but honestly, did not invest much time into this).

PR #1597, opened 2026-04-28
feat(pframes): EnrichmentRef + tableBuilder linker support

vadimpiven:
## Summary

- New `EnrichmentRef` (versioned `{__isEnrichment, hit, path?, qualifications?}`) accepted by `tableBuilder.addColumn`/`addColumns`. The build-table ephemeral registers the hit + every linker hop in the PFrame, calls `pframes.build-query.buildQuery`, and hands the resulting `SpecQueryJoinEntry` straight to a new internal `pt.p._queryEntry` — ptabler resolves the linker join natively, no node-by-node translation.
- Spec distiller preserves `pl7.app/isLinkerColumn`; without it ptabler's spec frame can't populate the linker index and `linkerJoin` silently degrades to inner join (`pframes-rs/.../frame.rs:90`).
- Per-step `qualifications` dropped from the typed shapes of `DiscoverColumnsLinkerStep`, `MatchVariant.path[]`, and `DiscoveredPColumn.path[]`. They were always-empty discovery metadata; `BuildQuery` already discarded them. `DiscoveredPColumnId`'s wire form keeps `qualifications: []` so existing IDs in block storage remain byte-compatible.
- New helper `enrichmentRefFromMatch(match, variant, refMap)` exported from `@platforma-sdk/model` for the model-side discovery → workflow plumbing.

WIP notes for resume across machines in `docs/enrichment-ref-tablebuilder-wip.md`.

## Status

The realistic-flow test rewrite (model-side `ColumnCollectionBuilder` discovery feeding `args.enrichment` into the workflow) is **in progress, not done** — pending a decision on where the linker columns come from for discovery (new fixture block vs. self-published from `filter-column-test`'s own workflow). See the WIP doc for the full open-questions list.

The current block test still exercises the `EnrichmentRef` end-to-end via inline synthesis; it passes against a backend running ptabler 1.16.0.

## Test plan

- [x] `pnpm -C lib/model/common test` — 112/112
- [x] `pnpm -C sdk/model test` — 366/366
- [x] `pnpm -C sdk/workflow-tengo test` — passing (5 new tengo tests)
- [x] `pnpm -C tests/workflow-tengo test src/pframes/build-query.test.ts src/pframes/table-builder.test.ts` — 18/18 (PL_ADDRESS)
- [x] `etc/blocks/filter-column-test/test` — 3/3, twice in a row (PL_ADDRESS)
- [ ] Multi-hop linker chain integration test
- [ ] EnrichmentRef-with-PlRefs in integration test
- [ ] Realistic flow via `ColumnCollectionBuilder` in args mapper

This PR introduces `EnrichmentRef` — a versioned descriptor for linker-reached enrichment columns — and wires it through `tableBuilder.addColumn/addColumns`, the `build-table` ephemeral template, and `pt.p._rawQueryEntry`, so ptabler resolves multi-hop linker joins natively without node-by-node translation. It also restructures `DatasetOption` / `PlDatasetSelector` to bundle discovered enrichments into a new `DatasetSelection` v-model, and fixes the spec distiller to preserve `pl7.app/isLinkerColumn` so ptabler can populate its linker index.

The two areas worth watching before broader adoption:
- `DatasetOption` shape and `PlDatasetSelector` v-model type are both breaking changes; any block that stores a `PrimaryRef` in args today will silently lose its dataset selection after upgrading — a migration read path or explicit migration guide is needed.
- `pt.p.linkerJoin`'s new annotation guard is skipped for unresolved futures, leaving a silent-degradation window if the linker column reaches ptabler without `pl7.app/isLinkerColumn`.
</details>

Safe to merge for new functionality; the breaking DatasetOption / PlDatasetSelector API change needs a migration path for existing blocks before production rollout.

All findings are P2 — no runtime correctness bugs in the new EnrichmentRef join path (verified by the inline block test). The DatasetOption/PlDatasetSelector breaking change is the most impactful but is an intended API evolution tracked in the changeset.

sdk/ui-vue/src/components/PlDatasetSelector/PlDatasetSelector.vue and sdk/model/src/components/PlDatasetSelector/dataset_selection.ts — both carry the breaking modelValue / DatasetOption shape change that will silently regress existing blocks.
</details>

| Filename | Overview |
|----------|----------|
| sdk/workflow-tengo/src/pframes/build-table.tpl.tengo | Core ephemeral template refactored: replaces node-by-node translateQueryEntry with pt.p._rawQueryEntry; adds path/step registration for EnrichmentRef linker hops; strips axesSpec: [] for path-bearing enrichments before passing to exportUtil |
| sdk/workflow-tengo/src/pframes/table-builder.lib.tengo | Adds EnrichmentRef / LabeledEnrichment support to addColumn/addColumns; serializes path steps as string-keyed map (serializer limitation); unconditionally marks all EnrichmentRefs as needing ctx even when fully pre-resolved |
| sdk/workflow-tengo/src/pt/index.lib.tengo | Adds _rawQueryEntry internal helper and a new pl7.app/isLinkerColumn annotation guard in linkerJoin; guard is correctly skipped for unresolved futures but leaves open a silent degradation path |
| sdk/workflow-tengo/src/pframes/spec.lib.tengo | Spec distiller now preserves pl7.app/isLinkerColumn annotation; all other annotations still discarded; new test confirms the behavior |
| lib/model/common/src/ref.ts | Adds EnrichmentRef, EnrichmentLinkerStep, LabeledEnrichmentRef types and createEnrichmentRef/isEnrichmentRef helpers; removes old DatasetOption (moved to sdk/model); well-tested |
| lib/model/common/src/drivers/pframe/spec/selectors.ts | Adds AxisQualification, MultiColumnSelector, matchMultiColumnSelector, multiColumnSelectorsToPredicate helpers; AxisQualification moved here from spec_driver.ts to break circular dependency |
| sdk/model/src/components/PlDatasetSelector/enrichment_discovery.ts | New file: findEnrichmentColumns + enrichmentVariantsToRefs; correctly propagates qualifications.forHit; filters out zero-hop and non-globally-addressable variants |
| sdk/model/src/components/PlDatasetSelector/dataset_selection.ts | New DatasetOption/DatasetSelection types; DatasetOption shape changed from Option& to {primary, filters?, enrichments?} — breaking change for all existing consumers of buildDatasetOptions |
| sdk/ui-vue/src/components/PlDatasetSelector/PlDatasetSelector.vue | v-model type changed to DatasetSelection; backward compat with PrimaryRef/PlRef modelValue removed; emitValue now looks up enrichments from options and bundles them into DatasetSelection |
| sdk/model/src/components/PlDatasetSelector/build_dataset_options.ts | Adds withEnrichments/enrichmentMaxHops options; uses multiColumnSelectorsToPredicate for primary selector; graceful degradation when collection or filters are absent |

</details>

</details>

`````markdown

---

### Issue 1 of 4
sdk/workflow-tengo/src/pt/index.lib.tengo:281-292
**Linker annotation guard skips futures silently**

The `pl7.app/isLinkerColumn` check is bypassed when `linkerSpec` is not yet a plain map (i.e. it's still a smart-object future). In that case no assertion fires, the column flows through to ptabler without the annotation, and ptabler silently downgrades the `linkerJoin` to an inner join (as noted in `pframes-rs/.../frame.rs:90`). The comment documents this intentionally, but it means the only guard against a misconfigured linker column is the spec-distiller test — there is no runtime catch if the annotation is absent on a column that arrives as a future.

Consider at minimum logging a warning when the spec is a future, or deferring the assertion to the `_rawQueryEntry` call site where the full join entry is already concrete.

### Issue 2 of 4
sdk/workflow-tengo/src/pframes/table-builder.lib.tengo:199-203
**Pre-resolved `EnrichmentRef` unconditionally requires ctx**

`isEnrichmentRef(query) → _hasUnresolved = true` is set regardless of whether `hit` and every `path[].linker` are already `{spec, data}` maps. A caller that constructs a fully pre-resolved `EnrichmentRef` without a workflow context will receive a panic from `build()` even though `resolveColumn` would never touch `_ctx` for those values.

In practice this is unlikely today, but the `TestAddColumnAcceptsEnrichmentRefWithResolvedHops` test confirms the construction path; if `.build()` is exercised without ctx the panic surfaces. The guard could be restricted to only set `_hasUnresolved` when at least one hop is a string or PlRef.

### Issue 3 of 4
sdk/model/src/components/PlDatasetSelector/dataset_selection.ts:1-37
**Breaking `DatasetOption` shape change not guarded by a version discriminant**

`DatasetOption` changed from `Option & {filters?}` (with `ref`/`label` at the top level) to `{primary: Option, filters?, enrichments?}`. Any existing caller that reads `option.ref` or `option.label` directly will silently get `undefined` at runtime. The type was also moved from `@milaboratories/pl-model-common` to `@platforma-sdk/model`, so import paths break.

If blocks still pass old-shaped `DatasetOption[]` to `PlDatasetSelector`, the dataset dropdown will render empty labels and broken refs. The changeset bump should call this out explicitly, and a migration guide (or a transitional `isLegacyDatasetOption` guard) would reduce the blast radius.

### Issue 4 of 4
sdk/ui-vue/src/components/PlDatasetSelector/PlDatasetSelector.vue:32
**Backward compatibility with `PrimaryRef | PlRef` modelValue removed**

The v-model type narrowed from `PrimaryRef | PlRef | undefined` to `DatasetSelection | undefined`. The two backward-compat tests were deleted. Any block that currently stores a `PrimaryRef` or `PlRef` in its args and passes it as `modelValue` will receive no dataset selection — `selectedDataset` returns `undefined` because `model.value?.primary.column` won't exist on those old shapes.

If existing blocks in production store `PrimaryRef` in block data, they will appear as "no dataset selected" after upgrading. Either add a migration read path (`isPrimaryRef(v) ? createDatasetSelection(v) : v`) or document that block data must be migrated.

`````

</details>

<sub>Reviews (2): Last reviewed commit: ["feat(pframes): EnrichmentRef + tableBuil..."](https://github.com/milaboratory/platforma/commit/36a87d1163876fe1e7a23b04955d84ab793a3c8a) | [Re-trigger Greptile](https://app.greptile.com/api/retrigger?id=30032734)</sub>

### DIFF
--- a/lib/model/common/src/ref.ts
+++ b/lib/model/common/src/ref.ts
@@ -1,4 +1,6 @@
 import { z } from "zod";
+import type { AxisQualification } from "./drivers/pframe/spec/selectors";
+import type { PObjectId } from "./pool/spec";
 
 export const PlRef = z
   .object({
@@ -125,10 +127,64 @@ export function createPrimaryRef(column: PlRef, filter?: PlRef): PrimaryRef {
 /** Block args accept either form — backward compatible. */
 export type DatasetInput = PlRef | PrimaryRef;
 
-/** Dataset option with optional filter choices. */
-export type DatasetOption = Option & {
-  readonly filters?: readonly Option[];
+/** Many-to-one linker hop. */
+export type EnrichmentLinkerStep = {
+  readonly type: "linker";
+  readonly linker: PObjectId;
 };
+
+/** Step from primary to hit. `path[0]` applied first. */
+export type EnrichmentStep = EnrichmentLinkerStep;
+
+/**
+ * Enrichment column reached via {@link path} linker hops; omitted means zero-hop.
+ * `hit` and `path[].linker` are global-form {@link PObjectId}s
+ * (`canonicalize({ __isRef: true, blockId, name })`) which the workflow decodes
+ * to a PlRef before resolving. For a prerun-sourced hop, supply a resolved
+ * `{spec, data}` map in place of the id — `tableBuilder` accepts both.
+ */
+export type EnrichmentRef = {
+  readonly __isEnrichment: "v1";
+  readonly hit: PObjectId;
+  readonly path?: readonly EnrichmentStep[];
+  readonly qualifications?: readonly AxisQualification[];
+};
+
+export function isEnrichmentRef(value: unknown): value is EnrichmentRef {
+  return (
+    typeof value === "object" &&
+    value !== null &&
+    (value as { __isEnrichment?: unknown }).__isEnrichment === "v1" &&
+    "hit" in value
+  );
+}
+
+export type CreateEnrichmentRefOptions = {
+  readonly path?: readonly EnrichmentStep[];
+  readonly qualifications?: readonly AxisQualification[];
+};
+
+export function createEnrichmentRef(
+  hit: PObjectId,
+  options: CreateEnrichmentRefOptions = {},
+): EnrichmentRef {
+  const { path, qualifications } = options;
+  return {
+    __isEnrichment: "v1",
+    hit,
+    ...(path !== undefined && path.length > 0 ? { path } : {}),
+    ...(qualifications !== undefined && qualifications.length > 0 ? { qualifications } : {}),
+  };
+}
+
+/** {@link EnrichmentRef} plus the export header label. */
+export type LabeledEnrichmentRef = {
+  readonly ref: EnrichmentRef;
+  readonly label: string;
+};
+
+/** Auto-attached enrichment payload; not user-picked. */
+export type LabeledEnrichmentRefs = readonly LabeledEnrichmentRef[];
 
 /** Compare two PlRefs and returns true if they are qual */
 export function plRefsEqual(ref1: PlRef, ref2: PlRef, ignoreEnrichments: boolean = false) {
--- a/sdk/model/src/components/PlDatasetSelector/build_dataset_options.ts
+++ b/sdk/model/src/components/PlDatasetSelector/build_dataset_options.ts
@@ -1,20 +1,25 @@
-import type {
-  DatasetOption,
-  Option,
-  PColumnSelector,
-  PObjectSpec,
-} from "@milaboratories/pl-model-common";
+import type { MultiColumnSelector, Option, PObjectSpec } from "@milaboratories/pl-model-common";
+import { multiColumnSelectorsToPredicate } from "@milaboratories/pl-model-common";
 import type { DeriveLabelsOptions } from "../../labels/derive_distinct_labels";
 import type { RenderCtxBase } from "../../render";
 import { ColumnCollectionBuilder } from "../../columns/column_collection_builder";
 import { collectCtxColumnSnapshotProviders } from "../../columns/ctx_column_sources";
+import type { DatasetOption } from "./dataset_selection";
 import { buildRefMap, filterMatchesToOptions, findFilterColumns } from "./filter_discovery";
+import { enrichmentVariantsToRefs, findEnrichmentColumns } from "./enrichment_discovery";
 
 export type BuildDatasetOptions = {
   /** Which result pool columns qualify as datasets. Defaults to all. */
-  selector?: PColumnSelector | PColumnSelector[] | ((spec: PObjectSpec) => boolean);
+  primary?: MultiColumnSelector | MultiColumnSelector[] | ((spec: PObjectSpec) => boolean);
   /** Formatting options for filter labels. */
   labelOptions?: DeriveLabelsOptions;
+  /**
+   * Enables enrichment discovery and filters hits attached to
+   * `DatasetOption.enrichments`. Use `() => true` to accept all; omit to disable.
+   */
+  withEnrichments?: MultiColumnSelector | MultiColumnSelector[] | ((spec: PObjectSpec) => boolean);
+  /** Maximum linker hops considered. Only used when `withEnrichments` is set. */
+  enrichmentMaxHops?: number;
 };
 
 /**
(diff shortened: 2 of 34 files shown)

### CODE REVIEW

File: sdk/model/src/components/PlDatasetSelector/build_dataset_options.ts [hunk 1]
```diff
@@ -1,20 +1,26 @@
...
 import type { RenderCtxBase } from "../../render";
 import { ColumnCollectionBuilder } from "../../columns/column_collection_builder";
 import { collectCtxColumnSnapshotProviders } from "../../columns/ctx_column_sources";
+import type { DatasetOption } from "./dataset_selection";
 import { buildRefMap, filterMatchesToOptions, findFilterColumns } from "./filter_discovery";
+import { enrichmentMatchesToRefs, findEnrichmentColumns } from "./enrichment_discovery";
 
 export type BuildDatasetOptions = {
   /** Which result pool columns qualify as datasets. Defaults to all. */
   selector?: PColumnSelector | PColumnSelector[] | ((spec: PObjectSpec) => boolean);
```
AStaroverov: can we use MultiColumnSelector instead PColumnSelector

sdk/model/src/components/PlDatasetSelector/enrichment_ref.ts
```diff
@@ -0,0 +1,39 @@
...
+
+export type EnrichmentRefFromMatchOptions = {
+  readonly match: ColumnMatch;
+  /** Specific variant to use — one match can have several reaching paths. */
+  readonly variant: MatchVariant;
+  /**
+   * Maps a discovered column's `PObjectId` to its upstream `PlRef`.
+   * Build once per render with `buildRefMap(ctx.resultPool.getSpecs().entries)`.
+   */
+  readonly refMap: ReadonlyMap<PObjectId, PlRef>;
```
AStaroverov: I think we can use PObjectId for workflow

sdk/model/src/components/PlDatasetSelector/enrichment_ref.ts
```diff
@@ -0,0 +1,39 @@
...
+  createEnrichmentRef,
+  type EnrichmentRef,
+  type EnrichmentStep,
+  type PlRef,
+  type PObjectId,
+} from "@milaboratories/pl-model-common";
+import type { ColumnMatch, MatchVariant } from "../../columns/column_collection_builder";
+
+export type EnrichmentRefFromMatchOptions = {
+  readonly match: ColumnMatch;
```
AStaroverov: use ColumnVariant

lib/model/common/src/ref.ts
```diff
@@ -125,11 +126,59 @@ export function createPrimaryRef(column: PlRef, filter?: PlRef): PrimaryRef {
...
+  return {
+    __isEnrichment: "v1",
+    hit,
+    ...(path !== undefined && path.length > 0 ? { path } : {}),
+    ...(qualifications !== undefined && qualifications.length > 0 ? { qualifications } : {}),
+  };
+}
+
+/** {@link EnrichmentRef} plus the export header label. */
+export type Enrichment = {
```
AStaroverov: LabeledEnrichmentRef

sdk/model/src/components/PlDatasetSelector/enrichment_discovery.ts
```diff
@@ -0,0 +1,80 @@
...
+ * Finds linker-reached hits attached to the anchor primary. Excludes
+ * subset, linker, and label hits (filters / structural / labels) and
+ * drops zero-hop matches (those are filters or the primary itself).
+ * Optional `predicate` further filters by hit-column spec.
+ */
+export function findEnrichmentColumns(
+  collection: AnchoredColumnCollection,
+  options?: { maxHops?: number; predicate?: (spec: PObjectSpec) => boolean },
+): ColumnMatch[] {
+  const matches = collection.findColumns({
```
AStaroverov: findColumnVariants can be mo suitable for this flow

AStaroverov (approved):

vadimpiven:
@greptileai

PR #1613, opened 2026-05-06
`pnpm run build:dev` deploys docker image to dev ECR registry making it available in AWS

blackcat:
`pnpm run build:dev` publishes software docker images to dev ECR registry


tools/package-builder/src/core/core.ts
```diff
@@ -368,21 +368,8 @@ export class Core {
...
       }
 
-      if (!artifacts.dockerArchitectures.includes(util.currentArch())) {
-        this.logger.log(
-          options?.strictPlatformMatching ? "debug" : "warn",
-          `Docker image generation was skipped because host architecture '${util.currentArch()}'` +
-            ` is currently not supported by Platforma Backend docker feature`,
-        );
-        continue;
-      }
```
DenKoren, line 378: I afraid this will make CI to build docker images twice: we run binary builds both on arm and amd architectures.

We still need to skip docker build in CI for incorrect architectures to not try building an image on windows host.

tools/package-builder/src/core/core.ts
```diff
@@ -714,9 +702,14 @@ export class Core {
...
-    const packagesToPublish = options?.ids ?? [
-      ...new Set([...this.buildablePackages.keys(), ...this.dockerPackages.keys()]),
-    ];
+    // Iterate only dockerPackages (matches buildDockerImages). Previously we
+    // also iterated buildablePackages.keys(), but for entrypoints with a python
+    // or conda binary that autogens a docker image, the binary key (e.g. "main")
+    // and the virtual docker key ("main:docker") both resolve to the same
+    // docker artifact via getArtifact's :docker fallback. That caused
+    // publishDockerImage to run twice — first push succeeded, second saw the
+    // tag in the registry and logged a misleading "Skipping push..." line.
```
DenKoren, line 723: reminder: did you test this?
I remember we had some pain with python software publishing as docker images.
Not sure was this saving us from the problem or not.

tools/package-builder/src/core/core.ts
```diff
@@ -731,21 +724,8 @@ export class Core {
...
       }
 
-      if (!artifacts.dockerArchitectures.includes(util.currentArch())) {
-        this.logger.log(
-          options?.strictPlatformMatching ? "debug" : "warn",
-          `Docker image generation was skipped because host architecture '${util.currentArch()}'` +
-            ` is currently not supported by Platforma Backend docker feature`,
-        );
-        continue;
-      }
```
DenKoren, line 741: Same story as above: CI will build images everywhere instead of doing this once on single runner.

DenKoren (approved):

tools/package-builder/src/core/core.ts
blackcat: Every time you asking this I feel internal trembling. Even if I already tested everything few times :D

tools/package-builder/src/core/core.ts
blackcat: fixed in 15f5a7b84

tools/package-builder/src/core/core.ts
blackcat: fixed in https://github.com/milaboratory/platforma/commit/15f5a7b842cca0e5728aa19b975af0fc4d53395e

PR #1615, opened 2026-05-06
MILAB-6145: WASM bundling — tengo-builder wasm artefact, :pframes-rs SDK lib, backend capability gate

blackcat:
## Summary

Two coherent pieces of the MILAB-6145 WASM-in-Tengo stream:

### 1. tengo-builder wasm artefact + `:pframes-rs` SDK lib (commit `0a3af0268`)

- `pl-model-backend`: `TemplateDataV3` gains optional `wasm` map mirroring the existing libs/software/assets shape. Wasm bytes are base64-encoded into `hashToSource[<sha256>]`, preserving the `hashToSource[hash] === stored value` invariant.
- `tengo-builder`: detects `assets.importWasm("@pkg:id")` (regex, like the other `import*` calls), resolves the bytes from each dependency's `package.json` `exports[*].wasm` condition (subpath `.` → id `main`, `./foo` → `foo`). Fallback: a `node_modules` walk for wasm-only packages with no JS entry point (otherwise `import.meta.resolve` rejects them).
- `@platforma-sdk/workflow-tengo`: new opt-in lib `:pframes-rs` wrapping `assets.importWasm("@milaboratories/pframes-rs-wasip2:main")`. Blocks that import this lib pull the upstream wasm into every template that imports it (transitively); blocks that don't, stay lean.

### 2. Backend capability gate (commit `8dcae58ff`)

- `BlockPackMeta` (pl-model-middle-layer): new optional `requiredCapabilities: string[]` field.
- `block-tools pack`: hardcodes `requiredCapabilities: ["wasm"]` for every block built with this SDK release. Coarse cut at the SDK version boundary — see [wasm-bundling.md](https://github.com/milaboratory/text/blob/main/work/projects/webassembly-libraries-tengo/wasm-bundling.md) for rationale.
- `pl-client`: surfaces `MaintenanceAPI.Ping.Response.capabilities` (proto field 9, added in the matching pl PR).
- `pl-middle-layer`: exposes `serverCapabilities` alongside `serverPlatform`.

Forward-compat: the new manifest field is optional + Zod strips unknown keys; protobuf wire format silently skips unknown fields. Old Desktops parse unchanged.

## Test plan

- [x] `pnpm turbo check` (full monorepo, pre-push hook) passes.
- [x] `tengo-builder` unit tests: 39/39 (2 new tests cover `assets.importWasm` detection + variable-id rejection).
- [x] `pl-model-middle-layer` unit tests: 4/4 (covers `requiredCapabilities` accept/round-trip/forward-compat).
- [x] **End-to-end verification against `blocks/rarefaction`**: template that imports `:pframes-rs` carries the full 1,725,128-byte wasm under `template.wasm["@milaboratories/pframes-rs-wasip2:main"]` in `main.plj.gz` (1.18 MB packed). Sibling template that doesn't import `:pframes-rs` carries zero wasm bytes (73 KB). `sha256(base64) === sourceHash` invariant holds.
- [ ] Manual: pack a real block and confirm `block-pack/manifest.json` contains `description.meta.requiredCapabilities = ["wasm"]`.

## Cross-repo deps

- pl backend: [milaboratory/pl#1798](https://github.com/milaboratory/pl/pull/1798) — `MaintenanceAPI.Ping.Response.capabilities`, `WasmV1` resource, controller wiring.
- Desktop app: [milaboratory/platforma-desktop-app#473](https://github.com/milaboratory/platforma-desktop-app/pull/473) — consumes the new SDK exports + gates block install on backend capability list.
- Block example: rarefaction PR (this branch on milaboratories/rarefaction) — demonstrates `import("@platforma-sdk/workflow-tengo:pframes-rs")`.

## Refs

- Project doc: [docs/text — webassembly-libraries-tengo](https://github.com/milaboratory/text/tree/main/work/projects/webassembly-libraries-tengo)
- Project: WebAssembly Libraries in Tengo (MILAB-6145)

This PR introduces WASM bundling support for the Tengo template pipeline and a backend capability gate so blocks that require the WASM runtime can declare that requirement and be gated at install time.

- **tengo-builder / `TemplateDataV3`**: `assets.importWasm(...)` calls are detected via the same regex pattern used for other `import*` calls; the resolved wasm bytes are base64-encoded and stored under `hashToSource`, keeping the existing hash-invariant intact. A new `:pframes-rs` opt-in lib in `workflow-tengo` bundles the 1.7 MB pframes-rs wasm only into templates that import it.
- **Capability gate**: `BlockPackMeta` gains `requiredCapabilities?: string[]`; `block-tools pack` stamps `[\"wasm\"]` when the compiled template tree carries any wasm sections (detected from the consolidated `main.plj.gz`); `pl-middle-layer` re-derives this at listing time for dev-v2 blocks and enforces it at both add-block call sites via `throwIfMissingServerCapabilities`; `pl-client` surfaces the new `capabilities` proto field 9 from the Ping response.
</details>

Safe to merge; the wasm gate is fail-closed (old backends return an empty capabilities list, blocking incompatible blocks), and the manifest field is optional so old Desktops parse unchanged.

The two copies of templateHasWasm (in build_dist.ts and required_capabilities.ts) are intentionally duplicated for a documented reason, but they share no shared contract — a future template format change applied to one copy but not the other would cause pack-time and install/listing-time capability detection to silently disagree. Everything else — the proto wiring, Zod schema change, the _bundleTag detection pattern, and the add-block guards — is well-structured and tested.

required_capabilities.ts and build_dist.ts for the duplicated templateHasWasm logic.
</details>

| Filename | Overview |
|----------|----------|
| tools/tengo-builder/src/compiler/compiler.ts | Adds wasm artifact type to ArtifactStore; minor typo in getWasm error message (missing closing quote). |
| tools/tengo-builder/src/compiler/main.ts | Adds WASM export resolution via package.json exports[*].wasm condition and a node_modules fallback walk; well-structured with clear conventions. |
| lib/node/pl-middle-layer/src/mutator/block-pack/required_capabilities.ts | New file deriving requiredCapabilities from parsed/raw workflow bytes; intentionally duplicates templateHasWasm from build_dist.ts — divergence risk if template format evolves. |
| tools/block-tools/src/v2/build_dist.ts | Adds workflowUsesWasm detection and stamps requiredCapabilities: ["wasm"] into the manifest when any wasm section is found; fail-safe (returns false on parse error). |
| lib/node/pl-middle-layer/src/middle_layer/project.ts | Adds throwIfMissingServerCapabilities guard at both add-block paths; clean and consistent with the existing feature-flag gate pattern. |
| lib/model/middle-layer/src/block_meta/block_meta.ts | Adds requiredCapabilities?: string[] and promotes schema to .passthrough() for forward-compat round-trips; well-tested and justified. |
| sdk/workflow-tengo/src/pframes-rs.lib.tengo | New opt-in Tengo library wrapping pframes-rs wasm; stub runtime returns clear errors; _bundleTag pattern is intentional for regex-based build-time detection. |
| lib/node/pl-client/src/proto-grpc/github.com/milaboratory/pl/plapi/plapiproto/api.ts | Adds proto field 9 (repeated string capabilities) to MaintenanceAPI_Ping_Response; serialisation/deserialisation is correct for unpacked strings. |
| lib/node/pl-middle-layer/src/block_registry/registry.ts | Re-derives requiredCapabilities for dev-v2 blocks at listing time to keep the UI gate consistent without requiring a pack step. |

</details>

</details>

```mermaid
flowchart TD
    A["pframes-rs.lib.tengo"] -->|"assets.importWasm detected by regex"| B["tengo-builder parseSource"]
    B --> C["ArtifactSource type=wasm base64 bytes"]
    C --> D["TengoTemplateCompiler addWasm"]
    D --> E["TemplateDataV3.wasm + hashToSource"]
    E --> F["main.plj.gz"]
    F --> G["buildBlockPackDist workflowUsesWasm"]
    G -->|"wasm found"| H["manifest.json requiredCapabilities wasm"]
    F --> I["BlockPackPreparer requiredCapabilitiesFromTemplate"]
    I --> J["BlockPackSpecPrepared.requiredCapabilities"]
    J --> K["throwIfMissingServerCapabilities"]
    L["pl backend Ping capabilities"] --> M["PlClient serverInfo.capabilities"]
    M --> K
    K -->|"missing"| N["Error install blocked"]
    K -->|"all present"| O["Block installed"]
```
</details>

`````markdown
Fix the following 2 code review issues. Work through them one at a time, proposing concise fixes.

---

### Issue 1 of 2
tools/tengo-builder/src/compiler/compiler.ts:263
The closing single quote is missing in the error string, so the diagnostic message reads `instead of 'wasm` rather than `instead of 'wasm'`. Compare the analogous check in `getAsset`/`getSoftware`, which use consistent quoting.

```suggestion
      throw new Error(`illegal artifact type: got ${name.type} instead of 'wasm'`);
```

### Issue 2 of 2
lib/node/pl-middle-layer/src/mutator/block-pack/required_capabilities.ts:5-20
`templateHasWasm` function divergence risk — the identical logic now lives in two packages (`build_dist.ts` and `required_capabilities.ts`). The comment explains the intent, but the two copies share no type contract: if a future template format change (e.g. renaming `templates` to something else, or adding another level) is applied to one but not the other, pack-time and install/listing-time capability derivation will silently disagree — a block that embeds wasm might pass the install gate, or a block that doesn’t might be incorrectly gated. Consider extracting this function (accepting `unknown`) into `@milaboratories/pl-model-backend` alongside `TemplateWasmDataV3`, so both call sites import the same implementation.

`````

</details>

<sub>Reviews (1): Last reviewed commit: ["MILAB-6145: dbolotin&#39;s comment https://g..."](https://github.com/milaboratory/platforma/commit/167e418987f66447c488430de968cdc5ac484f11) | [Re-trigger Greptile](https://app.greptile.com/api/retrigger?id=32140872)</sub>

### CODE REVIEW

File: lib/model/middle-layer/src/block_meta/block_meta.ts [hunk 1]
```diff
@@ -55,6 +55,21 @@ export function BlockPackMeta<
...
+     * Tokens are loose strings (not an enum) so future capabilities don't
+     * break Desktops that ship before each token is added — the same
+     * forward-compat property `BlockPackMeta` itself enjoys at the field
+     * level, applied recursively at the token level.
+     *
+     * Hardcoded by `@platforma-sdk/block-tools` to ["wasm"] for every
+     * block built with the SDK release that introduces WASM support.
+     * See docs/text/work/projects/webassembly-libraries-tengo/wasm-bundling.md.
+     */
+    requiredCapabilities: z.array(z.string()).optional(),
```
dbolotin: It is bad that this object don't have ` .passthrough()`, this means that if any block that will be published with previous version will strip-foo this field form all the blocks in the overview. Make sure you bump minor version of the `block-tools`, to make sure you trigger [this one](https://github.com/milaboratory/github-ci/blob/7e4702194c5ed7c1a762bbb42c98b2c27cb24e8e/.github/workflows/node-simple-pnpm.yaml#L687-L691).

lib/node/pl-mcp-server/README.md
```diff
@@ -30,14 +30,14 @@ Claude Code  ──MCP──▶  pl-mcp-server (worker)  ──IPC──▶  Des
...
-2. **Add the MCP server to Claude Code** by creating or editing `.claude/settings.json`:
+2. **Add the MCP server to Claude Code** by creating or editing `.claude/settings.json` (don't forget to change the MCP URL):
 
    ```json
    {
      "mcpServers": {
        "pl": {
          "type": "sse",
-         "url": "http://localhost:4200/mcp"
+         "url": "http://127.0.0.1:4200/<!!CHANGE ME!!>/mcp"
```
blackcat, line 40: Not part of the current task actually, just small doc fix.

lib/node/pl-middle-layer/src/mutator/block-pack/required_capabilities.ts
```diff
@@ -0,0 +1,47 @@
...
+
+/** Walks a parsed v3 template tree and returns true if any node carries a
+ * non-empty `wasm` map. Mirrors the build-time check in block-tools' pack
+ * (templateHasWasm in tools/block-tools/src/v2/build_dist.ts) so install
+ * time, listing time, and pack time stay in sync.
+ * Why do we need this code duplication?
+ * Only for development cycle. Block developer won't call do-pack on every
+ * workflow change, so manifest.json won't be populated with 'wasm'
+ * capability.
+ * */
```
blackcat: Pay attention to this part please ^^^ 
I didn't find better way.

blackcat: But it re-scans the code ONLY in dev lifecycle.

lib/model/middle-layer/src/block_meta/block_meta.ts
blackcat: passthrough added

lib/model/backend/src/template_data_v3.ts
```diff
@@ -59,3 +72,25 @@ export interface CompiledTemplateV3 {
...
+ *
+ * Single source of truth for "does this template require wasm?". Used by
+ * `block-tools` pack-time detection and by `pl-middle-layer` install-time
+ * gating — keeping the two in sync avoids a class of bug where a block
+ * embedding wasm could slip past the install gate (or vice versa) after
+ * a template-format change applied to only one copy.
+ *
+ * Accepts `unknown` so callers can pass parsed JSON without a prior
+ * type assertion. */
+export function templateHasWasm(tpl: unknown): boolean {
```
kevindetry-milaboratories: Maybe a nitpick, but this feels like reinventing zod.

lib/node/pl-middle-layer/src/middle_layer/project.ts
```diff
@@ -985,3 +988,26 @@ function convertErrorsToStrings(
...
+ * the UI (`AddBlockModal`) and direct middle-layer callers (the MCP server's
+ * `add_block` tool, programmatic block installs, tests) reject incompatible
+ * blocks consistently. */
+function throwIfMissingServerCapabilities(
+  pl: PlClient,
+  requiredCapabilities: readonly string[] | undefined,
+): void {
+  if (!requiredCapabilities || requiredCapabilities.length === 0) return;
+  const advertised = (pl.serverInfo.capabilities ?? []) as readonly string[];
+  const missing = requiredCapabilities.filter((c) => !advertised.includes(c));
```
kevindetry-milaboratories: No need to create an array, just use `.some`.

sdk/workflow-tengo/package.json
```diff
@@ -26,6 +26,7 @@
     "do-pack": "rm -f *.tgz && pnpm pack && mv *.tgz package.tgz"
   },
   "dependencies": {
+    "@milaboratories/pframes-rs-wasip2": "1.1.34",
```
kevindetry-milaboratories: Probably also should be in catalog?

tools/tengo-builder/src/compiler/main.ts
```diff
@@ -79,15 +87,36 @@ function resolvePackageJsonPackage(root: string, depPackageName: string): string
...
+    return undefined;
+  }
+  throw new Error(
+    `Can't resolve package.json for package ${depPackageName ?? "."} relative to ${root}`,
+  );
+}
+
+function findInNodeModules(root: string, depPackageName: string): string | undefined {
+  let dir = root;
+  for (let i = 0; i < 12; i++) {
```
kevindetry-milaboratories: This `12` thing looks rather suspicious.

lib/model/backend/src/template_data_v3.ts
```diff
@@ -59,3 +72,20 @@ export interface CompiledTemplateV3 {
...
+/** Returns true if a v3 template tree (or any nested fragment) carries
+ * a non-empty `wasm` map. Recursion descends into `templates`, so wasm
+ * deps in transitively-imported sub-templates are detected too.
+ *
+ * Single source of truth for "does this template require wasm?". Used by
+ * `block-tools` pack-time detection and by `pl-middle-layer` install-time
+ * gating — keeping the two in sync avoids a class of bug where a block
+ * embedding wasm could slip past the install gate (or vice versa) after
+ * a template-format change applied to only one copy. */
+export function templateHasWasm(tpl: TemplateDataV3): boolean {
```
DenKoren: Maybe it is better to expose a metadata attribute to CompiledTemplateV3, that exposes features required by template? Put them into the pack at compile time, extract them on block-tools level instead of recursive analysis of the whole template dependency tree. Then just match strings in that required features list against list presented by backend. List all known features in single TS file that is shared across pl-client, model and tengo-builder.

Then you 'teleport' requirements from workflow to block-tools side, like you do from block-tools level to ML. This will concentrate the logic of definition 'what to require for workflow' on package-builder side.
block-tools just merge its own requirements (if any) with workflow requirements (if any) and pass combination to a higher level.

tools/tengo-builder/src/compiler/compiler.ts
DenKoren, line 1: What happens if we build block that _has_ wasm references with old tengo-builder that does not support wasm embedding (just a version prior this one)?

My intuition says 'it will produce valid artifact, block-tools will accept it, but the runtime will fail'.

I think we need some sort of dependency requirement on any level. Either block-tools somehow detect that workflow was built by old tengo-builder, or we force fresh version of tengo-builder from CI like we do for block-tools. Or any other sort of sanity check, that ensures workflow code is built the right way.

tools/tengo-builder/src/compiler/main.ts
```diff
@@ -14,6 +14,13 @@ import type winston from "winston";
 import { tryResolve, tryResolveOrError } from "@milaboratories/resolve-helper";
 import { serializeTemplate } from "@milaboratories/pl-model-backend";
 
+// Convention: wasm component packages declare their wasm under a "wasm" condition
+// in package.json `exports`. Subpath "." maps to the tengo id "main"; subpath
+// "./foo" maps to id "foo". Other shapes of `exports` (string-valued, condition-only)
+// don't produce wasm artifacts.
+const wasmExportCondition = "wasm";
+const defaultWasmExportId = "main";
```
DenKoren, line 27: Is this convention standard, or it's something we introduced?

tools/tengo-builder/src/compiler/main.ts
```diff
@@ -313,6 +360,104 @@ function loadAssetsFromDir(
...
+  // with the existing `hashToSource[hash] === stored value` invariant.
+  const bytes = fs.readFileSync(file);
+  const source = bytes.toString("base64");
+  const wasm = new ArtifactSource(mode, fullName, getSha256(source), source, file, [], []);
+
+  logger.debug(`Adding dependency ${fullNameToString(fullName)} from ${file}`);
+  compiler.addWasm(wasm);
+}
+
+function collectWasmExportPaths(packageRoot: string): WasmExport[] {
```
DenKoren, line 476: Please, document why we need such curious way of wasm detection.
Without hands in task it is very hard to understand the motivation.

It is a bit hard to get what's happening from the code
Having an example of desired shape you look in package.json inside this function would be helpful.
As long as you don't want to use zod here, preferring raw objects traversal, at least comment in function's header is required to be clear.

tools/block-tools/src/v2/build_dist.test.ts
```diff
@@ -14,3 +15,70 @@ test.skip("create dist test", async () => {
   const manifest = await buildBlockPackDist(description, distPath);
   console.dir(manifest, { depth: 5 });
 });
+
+describe("templateHasWasm", () => {
```
DenKoren: It's strange to have in block-tools tests for function for pl-model-backend, that rely on type structure defined in the same package. Move them closer to to a place where all the logic lives.

lib/node/pl-middle-layer/src/middle_layer/project.ts
```diff
@@ -985,3 +988,27 @@ function convertErrorsToStrings(
...
+  if (!requiredCapabilities || requiredCapabilities.length === 0) return;
+  const advertised = (pl.serverInfo.capabilities ?? []) as readonly string[];
+  const filterFn = (c: string) => !advertised.includes(c);
+  if (!requiredCapabilities.some(filterFn)) return;
+  const missing = requiredCapabilities.filter(filterFn);
+  throw new Error(
+    `Block cannot be added: connected backend does not advertise capabilities ` +
+      `${JSON.stringify(missing)}. ` +
+      `Backend advertises ${JSON.stringify(advertised)}. ` +
+      `Upgrade the backend or remove the requirement from the block manifest.`,
```
DenKoren, line 1012: Is this an error end-user's would see? How they could 'remove the requirement from the block manifest'?

```suggestion
      `Upgrade the backend or use older version of a block.`,
```

lib/node/pl-middle-layer/src/block_registry/registry.ts
```diff
@@ -207,9 +208,24 @@ export class BlockPackRegistry {
...
+              // Re-derive here so the UI can raise a modal on "Add to Project" after
+              // the user clicks — matches the install-time check in
+              // mutator/block-pack/block_pack.ts.
+              if (meta.requiredCapabilities === undefined) {
+                const workflowBytes = await fs.promises.readFile(
+                  v2Description.components.workflow.main.file,
+                );
+                const derived = deriveRequiredCapabilities(workflowBytes);
+                if (derived !== undefined) meta.requiredCapabilities = derived;
+              }
```
DenKoren, line 228: See my question for tengo-builder updates. Looks like this does not protect us from anything unless we force use correct version of tengo-builder.
If we do - then 
- capabilities list would exist for modern blocks that can have wasm
- older blocks are guaranteed to have no wasm deps anyway.

lib/node/pl-middle-layer/src/middle_layer/project.ts
blackcat: No! The end user will see nice desktop recommendation https://platforma-bio.slack.com/archives/C0AMSHEPSSK/p1778845021044939?thread_ts=1778775383.699669&cid=C0AMSHEPSSK

This error is for block developers that try to do something from middle-layer level. The only way to see this error now _without hacking_ our code is MCP Servers add_block_tool.

tools/tengo-builder/src/compiler/main.ts
blackcat: We introduced it.

tools/tengo-builder/src/compiler/compiler.ts
blackcat: Same as dbolotin mentioned before https://github.com/milaboratory/platforma/pull/1615#discussion_r3197756697

Yes, this is the same pattern as the block-tools defense above. Old tengo-builder doesn't know the assets.importWasm() regex, so:

- The compiled template gets no wasm section.
- block-tools (even the new one) finds no wasm bytes to flag → no requiredCapabilities.
- The block installs anywhere and fails at workflow runtime.

BUT

The tengo-builder bump in this PR's changeset is already minor, so we can ride the same require-latest mechanism — proposing a follow-up to milaboratory/github-ci's node-simple-pnpm.yaml to add @platforma-sdk/tengo-builder alongside @platforma-sdk/block-tools in the require-latest step. Same defense, same gap: it only catches consumers publishing through that CI workflow; locally-built and locally-published artifacts still bypass it.

DenKoren: What do you think of making major version bump for tengo-builder?
Not all the features in tengo-builder are such dangerous and have such consequences on the full flow.
Actually, this is the first one with such scenario in the last 2 years.

Requiring devs to keep latest minor in all workflows may be annoying and over-kill.
One block-tools already disturbs enough.

We actually have forward-compatibility issue here between workflow-tengo and tengo-builder. There's no right mechanism to solve this and it is actually not covered by semver as I know. I think it is OK to step away from semver in this particular case and bump major, forcing it in CI.

The correct place to  solve a problem is to _require_ fresh tengo-buidler from fresh workflow-tengo, but we don't have such instruments and I don't think we need to develop them now, honestly.

DenKoren (approved):

PR #1649, opened 2026-05-21
Add pl-healthcheck package; use it for backend readiness check

DenKoren:
Introduce @milaboratories/pl-healthcheck — a thin TS client for the standard grpc.health.v1.Health service. Replace the UnauthenticatedPlClient ping in LocalPl.isAlive with a standard health Check RPC and tighten the isAlive contract to throw on timeout instead of returning a boolean.

This PR introduces `@milaboratories/pl-healthcheck`, a thin TypeScript client for the standard `grpc.health.v1.Health` protocol, and replaces the previous TCP-probe + boolean-return `isAlive` in `LocalPl` with a polling loop that throws on timeout and uses the new gRPC health check.

- **New `pl-healthcheck` package** (`src/client.ts`) exposes `PlHealthClient` with a single-shot `check()` and a convenience `waitServing()` poller, backed by `@protobuf-ts`-generated stubs from the canonical upstream proto.
- **`LocalPl.isAlive`** is redesigned to loop until the process is alive and the health check returns `SERVING`, throwing a descriptive error on timeout; the old TCP-probe helper is removed.
- **`simple.test.ts`** adopts the new API, but passes `pollIntervalMs: 30`, which inadvertently sets the per-RPC gRPC timeout to 30 ms — a concern noted in the review comments.
</details>

Mostly safe but the poll-interval-as-RPC-timeout conflation can cause health checks to never succeed with short poll intervals.

The new gRPC health-check approach is a solid improvement over the old TCP probe. The main concern is that isAlive feeds pollMs directly into PlHealthClient.build({ defaultRequestTimeout: pollMs }), so callers that pass a small poll interval (the integration test uses 30 ms) also impose a 30 ms deadline on each gRPC call. Under any non-trivial load or on first channel establishment, 30 ms is routinely not enough, meaning the health check can time out repeatedly even against a healthy backend.

lib/node/pl-deployments/src/local/pl.ts — the PlHealthClient construction where pollMs is reused as the RPC timeout.
</details>

| Filename | Overview |
|----------|----------|
| lib/node/pl-deployments/src/local/pl.ts | isAlive rewritten to poll with gRPC health checks; has a bug where pollIntervalMs is also used as the per-RPC timeout, making short poll intervals produce unreliable health checks. |
| lib/node/pl-healthcheck/src/client.ts | New PlHealthClient wrapping gRPC Health v1; clean implementation with both single-shot check() and waitServing() helpers. |
| lib/node/pl-healthcheck/package.json | New package manifest with correct dependencies on @grpc/grpc-js and @protobuf-ts/* from the workspace catalog. |
| tests/config-local-ml-integration/src/simple.test.ts | Caller updated from manual poll loop to isAlive({pollIntervalMs:30}); however passing 30ms as the poll interval also inadvertently sets the gRPC RPC timeout to 30ms. |
| .changeset/smooth-kiwis-watch.md | Changeset for the new pl-healthcheck package; has a minor typo ("Poper" instead of "Proper"). |
| lib/node/pl-healthcheck/generate-grpc.sh | Shell script for regenerating protobuf-ts output; uses safe bash options and correctly parameterizes tool paths. |

</details>

</details>

```mermaid
sequenceDiagram
    participant Caller
    participant LocalPl
    participant OS as OS (isProcessAlive)
    participant PlHealthClient
    participant Backend as Platforma Backend (gRPC)

    Caller->>LocalPl: "isAlive({ timeoutMs, pollIntervalMs })"
    loop until SERVING or deadline
        LocalPl->>LocalPl: sleep(pollMs)
        LocalPl->>OS: isProcessAlive(pid)
        OS-->>LocalPl: true/false
        alt process not alive
            LocalPl->>LocalPl: continue
        else grpcPort undefined
            LocalPl-->>Caller: return (void)
        else grpcPort defined
            LocalPl->>PlHealthClient: check()
            PlHealthClient->>Backend: grpc.health.v1.Health/Check
            Backend-->>PlHealthClient: ServingStatus
            PlHealthClient-->>LocalPl: SERVING / other / throw
            alt SERVING
                LocalPl-->>Caller: return (void)
            else not yet ready
                LocalPl->>LocalPl: continue loop
            end
        end
    end
    LocalPl-->>Caller: throw Error("Timeout")
```
</details>

<details open><summary><h3>Comments Outside Diff (2)</h3></summary>

1. `lib/node/pl-deployments/src/local/pl.ts`, line 94-98 ([link](https://github.com/milaboratory/platforma/blob/4823fb5c3b22ba26ee840e69e67b8c64e90c9c9f/lib/node/pl-deployments/src/local/pl.ts#L94-L98)) 

   <a href="#">P1</a> **RPC timeout set to poll interval — too short in test usage**

   `PlHealthClient` is built with `defaultRequestTimeout: pollMs`, but the test calls `isAlive({ pollIntervalMs: 30 })`, meaning each health-check gRPC call gets a **30 ms** deadline. A local gRPC round-trip is typically a few ms but can easily spike past 30 ms under load or on first connection establishment, causing every health-check attempt to time out even when the backend is fully healthy. `pollIntervalMs` (how often to retry) and the per-RPC deadline (how long each call may block) are independent concerns; collapsing them into one value breaks either fast-polling or reliable detection.

   `````markdown
   Path: lib/node/pl-deployments/src/local/pl.ts
   Line: 94-98

   Comment:
   **RPC timeout set to poll interval — too short in test usage**

   `PlHealthClient` is built with `defaultRequestTimeout: pollMs`, but the test calls `isAlive({ pollIntervalMs: 30 })`, meaning each health-check gRPC call gets a **30 ms** deadline. A local gRPC round-trip is typically a few ms but can easily spike past 30 ms under load or on first connection establishment, causing every health-check attempt to time out even when the backend is fully healthy. `pollIntervalMs` (how often to retry) and the per-RPC deadline (how long each call may block) are independent concerns; collapsing them into one value breaks either fast-polling or reliable detection.

   `````
   </details>

2. `lib/node/pl-deployments/src/local/pl.ts`, line 75-106 ([link](https://github.com/milaboratory/platforma/blob/4823fb5c3b22ba26ee840e69e67b8c64e90c9c9f/lib/node/pl-deployments/src/local/pl.ts#L75-L106)) 

   <a href="#">P2</a> **Unconditional sleep before the first health check**

   The loop always executes `await sleep(pollMs)` before any process or health check. If the backend is already running and healthy when `isAlive` is called (e.g., in a restart or reconnect scenario), the call will still stall for at least one `pollMs` (500 ms by default) before returning. Moving the sleep to the bottom of the loop (or using a `do … while` structure) would allow immediate return when the backend is ready on the very first poll.

   `````markdown
   Path: lib/node/pl-deployments/src/local/pl.ts
   Line: 75-106

   Comment:
   **Unconditional sleep before the first health check**

   The loop always executes `await sleep(pollMs)` before any process or health check. If the backend is already running and healthy when `isAlive` is called (e.g., in a restart or reconnect scenario), the call will still stall for at least one `pollMs` (500 ms by default) before returning. Moving the sleep to the bottom of the loop (or using a `do … while` structure) would allow immediate return when the backend is ready on the very first poll.

   `````
   </details>

</details>

`````markdown

---

### Issue 1 of 4
lib/node/pl-deployments/src/local/pl.ts:94-98
**RPC timeout set to poll interval — too short in test usage**

`PlHealthClient` is built with `defaultRequestTimeout: pollMs`, but the test calls `isAlive({ pollIntervalMs: 30 })`, meaning each health-check gRPC call gets a **30 ms** deadline. A local gRPC round-trip is typically a few ms but can easily spike past 30 ms under load or on first connection establishment, causing every health-check attempt to time out even when the backend is fully healthy. `pollIntervalMs` (how often to retry) and the per-RPC deadline (how long each call may block) are independent concerns; collapsing them into one value breaks either fast-polling or reliable detection.

### Issue 2 of 4
lib/node/pl-deployments/src/local/pl.ts:179-181
The JSDoc for `grpcPort` still says "MaintenanceAPI.Ping RPC", but the implementation now uses the standard gRPC Health v1 `Check` RPC via `PlHealthClient`. The description should reflect the actual protocol used.

```suggestion
   * When set, {@link LocalPl.isAlive} additionally issues a gRPC Health v1
   * `Check` RPC against this port so callers can distinguish a process that
   * is alive but still initializing from one actually serving gRPC.
```

### Issue 3 of 4
.changeset/smooth-kiwis-watch.md:6
Typo in the changeset description — "Poper" should be "Proper".

```suggestion
Proper platforma backend readiness checks
```

### Issue 4 of 4
lib/node/pl-deployments/src/local/pl.ts:75-106
**Unconditional sleep before the first health check**

The loop always executes `await sleep(pollMs)` before any process or health check. If the backend is already running and healthy when `isAlive` is called (e.g., in a restart or reconnect scenario), the call will still stall for at least one `pollMs` (500 ms by default) before returning. Moving the sleep to the bottom of the loop (or using a `do … while` structure) would allow immediate return when the backend is ready on the very first poll.


`````

</details>

<sub>Reviews (1): Last reviewed commit: ["Add pl-healthcheck package; use it for b..."](https://github.com/milaboratory/platforma/commit/4823fb5c3b22ba26ee840e69e67b8c64e90c9c9f) | [Re-trigger Greptile](https://app.greptile.com/api/retrigger?id=33200830)</sub>

### CODE REVIEW

File: lib/node/pl-healthcheck/src/client.ts [hunk 1]
```diff
@@ -0,0 +1,79 @@
...
+  async check(service: string = "", timeoutMs?: number): Promise<ServingStatus> {
+    const { response } = await this.client.check(
+      { service },
+      { timeout: timeoutMs ?? this.defaultTimeout },
+    );
+    return response.status;
+  }
+
+  /** Polls Check until SERVING or deadline. Returns true on SERVING, false on timeout. */
+  async waitServing(
```
AStaroverov: after serving it cannot be fallen?

AStaroverov: if it can, I would like to see classic listener pattern for every status

lib/node/pl-deployments/src/local/pl.ts
```diff
@@ -95,10 +96,91 @@ export class LocalPl {
...
+   *
+   * Calling this method is an explicit statement that the caller wants a real
+   * readiness signal; if `grpcPort` is undefined we throw rather than silently
+   * degrade to a process-only check (use {@link isAlive} for that).
+   */
+  async isReady(opts?: {
+    grpcPort?: number;
+    timeoutMs?: number;
+    pollIntervalMs?: number;
+  }): Promise<void> {
```
AStaroverov, line 129: why `isReady` return `void`? `boolean ` will be more suitable  return type

AStaroverov: or we should change name to `awaitingReadiness` or something like that

lib/node/pl-healthcheck/src/client.ts
AStaroverov: it's dead code, why we implement it?

lib/node/pl-deployments/src/local/pl.ts
```diff
@@ -95,10 +96,91 @@ export class LocalPl {
...
+      try {
+        if (healthClient === undefined) {
+          healthClient = PlHealthClient.build({
+            address: `127.0.0.1:${grpcPort}`,
+            defaultRequestTimeout: pollMs,
+          });
+        }
+
+        const status = await healthClient.check();
+        if (status === ServingStatus.SERVING) return;
```
AStaroverov, line 174: Are sure that we need a class here? `async function getPlHealthCLientStatus(): Promise<...>` will be enough

lib/node/pl-deployments/src/local/pl.ts
DenKoren: This contract is bad.
ConnectLocal in desktop app already has `await plLocal.isAlive();`. When you read the code, you read it like 'instance is alive after this line of code'. But in reality it can be in any state.

I think it is better to throw an exception to clearly show there's a problem, instead of leaving this situation handling to each caller.

lib/node/pl-healthcheck/src/client.ts
DenKoren: Dropped.

lib/node/pl-deployments/src/local/pl.ts
DenKoren: Honestly, I did not thought much on this.
Just repeated the pattern we have in pl-client for working with gRPC client.

After considering a while, I still don't see anything bad of its existence.
The choice between two here looks more like a matter of individual preference.

lib/node/pl-healthcheck/src/client.ts
DenKoren: Out of this PR scope, but answering your question: we did much precautions in backend to make this impossible, but honest answer would be yes. :(

At least once we observed the situation, when process was hang and stayed alive, responding to API, but only half-working. After that we created this status-checker.

It ensures all inner components are OK. If any component fails - checker shows this. Kubernetes automatically restarts pod for us in this case.

lib/node/pl-deployments/src/local/pl.ts
AStaroverov: not really, Class you want for internal hidden state. Pure function it's best thing in programming, so if you can implement it on pure function please do it.

DenKoren: Ok.
GRPC client has cost of initialisation, underlying connection manager and so on.
Paying it for each check of service status is not good in general sense. But here it is not on hot path anyway and will affect only start of a service once. Does not matter much in particular case.

So, I still think it is sort of personal preference. :)
Done.

AStaroverov (approved):

PR #1656, opened 2026-05-25
Milab-6145 WASM monorepo tests

blackcat:

This PR wires up end-to-end WASM component support in the Tengo template system: `tengo-builder` learns to bundle `.wasm` artifacts into template packs, `pl-middle-layer` renders them as `Wasm/1` resources, and the `workflow-tengo` SDK exposes `assets.importWasm`. The `pframes-rs` wrapper is upgraded from a stub to a live implementation, and a comprehensive integration test suite is added.

- `BackendCapability` is promoted from a string union type to a const-object + type pair; `PlWasmV1` namespace and `wasmField`/`wasmPrefix` helpers are added to the model layer.
- `direct_template_loader_v3.ts` adds a `WasmRenderer` and threads wasm fields into `TemplateRenderer`, following the existing `LibRenderer`/`SoftwareInfoRenderer` patterns.
- The `wasm-fixture` workspace package bundles a pre-built WASI p2 `table_component.wasm` and a `sync-wasm-fixture.sh` script for refreshing it from the `milaboratory/pl` repo.
</details>

The production-path changes (model layer, loader, SDK wrapper) are small, well-contained, and follow existing patterns. The bulk of the diff is new integration tests that are skipped on backends without WASM support.

All library and SDK changes are additive and follow established patterns. The pframes-rs upgrade removes a stub and replaces it with a straightforward live implementation. The few rough edges are confined to test templates and the sync script, and none affect correctness of the production code paths.

A handful of test templates (state-isolation.tpl.tengo, trap-then-fresh.tpl.tengo, bad-json.tpl.tengo) skip the is_error guard after importWasm, and store-budget.tpl.tengo returns raw integers rather than strings. The default SSH origin in sync-wasm-fixture.sh may cause issues in CI.
</details>

| Filename | Overview |
|----------|----------|
| lib/node/pl-client/src/core/capabilities.ts | Converts BackendCapability from a string union type to a const enum object + type, enabling value-level use without breaking existing type consumers. |
| lib/model/backend/src/template_resources_v1.ts | Adds PlWasmV1 namespace with resource type, constants, and fromV3Data factory; extends PlTemplateV1 with wasmPrefix/wasmField helpers. |
| lib/node/pl-middle-layer/src/mutator/template/direct_template_loader_v3.ts | Adds WasmRenderer (cache-key + render) and wires wasm fields into TemplateRenderer, consistent with LibRenderer and SoftwareInfoRenderer patterns. |
| sdk/workflow-tengo/src/assets/index.lib.tengo | Adds importWasm — a thin wrapper over plapi.loadWasm that serves as both the build-time regex anchor and runtime entry point. |
| sdk/workflow-tengo/src/pframes-rs.lib.tengo | Replaces stub _bundleTag/_runtimeUnavailable with a real _load calling assets.importWasm; fixes json.encode → string() coercion; updates error-branch checks from is_undefined to is_error. |
| tools/tengo-builder/src/compiler/compiler.ts | Switches hard-coded "wasm:v1" string to BackendCapability.WasmV1 constant; no logic changes. |
| tests/workflow-tengo/src/wasm/state-isolation.tpl.tengo | Missing is_error guard on wasmA/wasmB before accessing .exports — diverges from the pattern used in every other template in this PR. |
| tests/workflow-tengo/src/wasm/store-budget.tpl.tengo | Returns raw Tengo integers for firstErrAt/totalErrs while similar templates wrap counts in string(); test assertions expect JS numbers, which may be fragile. |
| tests/workflow-tengo/wasm-fixture/sync-wasm-fixture.sh | Well-structured sparse-clone script; defaults to SSH origin which fails in CI environments without SSH key setup. |
| tests/workflow-tengo/src/wasm/wasm.test.ts | End-to-end test covering happy path, result.Err arm, and guest trap; correctly gates on BackendCapability.WasmV1. |
| tests/workflow-tengo/wasm-fixture/package.json | Defines the wasm export condition mapping ./:main → ./table_component.wasm; correctly scoped as private workspace package. |

</details>

</details>

```mermaid
sequenceDiagram
    participant TB as tengo-builder
    participant ML as pl-middle-layer (loader)
    participant BE as pl backend
    participant TNG as Tengo template

    TB->>TB: regex-scan source for assets.importWasm("pkg:id")
    TB->>TB: resolve pkg exports[wasm] → read .wasm bytes → base64-encode
    TB->>TB: "embed in CompiledTemplateV3.wasm map + set requiredCapabilities=[wasm:v1]"

    ML->>ML: WasmRenderer.updateCacheKey (hash name+version+bytes)
    ML->>BE: "tx.createValue(Wasm/1, JSON{Name,Version,Runtime,Code,DefaultMemoryLimit})"
    ML->>BE: tx.setField(tplRef, "wasm/alias", wasmRef)

    BE->>BE: precompile + cache component bytes
    TNG->>BE: plapi.loadWasm("pkg:id") allocate Store + Instance
    BE-->>TNG: "sandboxed component handle {exports:{...}}"
    TNG->>TNG: handle.exports[iface][resource][method](argsJson)
    BE-->>TNG: resultJson or Tengo error (Err arm / trap)
```
</details>

`````markdown
Fix the following 3 code review issues. Work through them one at a time, proposing concise fixes.

---

### Issue 1 of 3
tests/workflow-tengo/src/wasm/state-isolation.tpl.tengo:36-40
**Missing `is_error` check before accessing `.exports`**

`wasmA` and `wasmB` are accessed unconditionally at `.exports["demo:tables/api"]`. If `assets.importWasm` returns a Tengo error (e.g., the binary is missing or a version mismatch occurs after the capability check passes), the subsequent `.exports` key access will produce a `nil/undefined` value rather than a diagnostic string, and the template will fail with a cryptic Tengo runtime error.

The same pattern is absent in `trap-then-fresh.tpl.tengo` (lines 19–20 and 23–24) and inside the `probe` closure in `bad-json.tpl.tengo`. Every other template in this PR gates on `if is_error(wasm) { return ... }` immediately after the import call — these three are inconsistent with that idiom and with the documented API contract ("check with `is_error(...)` before use").

### Issue 2 of 3
tests/workflow-tengo/src/wasm/store-budget.tpl.tengo:32-36
**Raw integer outputs may not round-trip through `getDataAsJson()`**

`firstErrAt` and `totalErrs` are returned as Tengo integers while `near-budget.tpl.tengo` (which counts the same kind of per-render tally) explicitly wraps its counter in `string()` with the comment "Wrap in string() so every output is a JSON-string resource." The corresponding test then asserts `expect(firstErrAt).toBe(64)` (JS number), relying on the framework storing and re-parsing an integer resource rather than a string resource. If the template output layer only produces JSON-string resources for Tengo scalars, `getDataAsJson()` may return `"64"` (string) and the strict `toBe(64)` assertion will fail.

### Issue 3 of 3
tests/workflow-tengo/wasm-fixture/sync-wasm-fixture.sh:21
The default `SYNC_ORIGIN` uses the SSH URL (`git@github.com:...`), which fails in CI runners that don't have SSH keys configured for GitHub. Using the HTTPS URL as the default is safer and still allows SSH overrides from the calling environment.

```suggestion
: "${SYNC_ORIGIN:="https://github.com/milaboratory/pl.git"}"
```

`````

</details>

<sub>Reviews (2): Last reviewed commit: ["MILAB-6145: complete WASM integration te..."](https://github.com/milaboratory/platforma/commit/7821f80297f199cd50aab2abc20e04c4d28908c2) | [Re-trigger Greptile](https://app.greptile.com/api/retrigger?id=33838810)</sub>

### CODE REVIEW

File: lib/node/pl-client/src/core/capabilities.ts [hunk 1]
```diff
@@ -15,7 +15,12 @@
  *   - `@platforma-sdk/block-tools`       — copies it onto the published manifest
  *   - `@milaboratories/pl-middle-layer`  — checks it at install time
  */
-export type BackendCapability = "auth:v2" | "treeFilter:v2" | "wasm:v1";
+export const BackendCapability = {
+  AuthV2: "auth:v2",
+  TreeFilterV2: "treeFilter:v2",
+  WasmV1: "wasm:v1",
+} as const;
+export type BackendCapability = (typeof BackendCapability)[keyof typeof BackendCapability];
```
DenKoren: nit:
Hah. Dima caught me exactly on similar thing.
This works and does what you want here, but in practice it is almost useless complication.

Old way still required you to use correct values everywhere where type is specified. The benefit of it is you immediately see the actual value in the code, not a name of a constant that you need to resolve while reading.

If I missed some useful value of having const object here - let's leave it as is. If not - maybe we should leave older way for simplicity

tests/workflow-tengo/src/wasm/echo-roundtrip.tpl.tengo
```diff
@@ -0,0 +1,38 @@
...
+	api := wasm["demo:tables/api"]
+	return string(api.table.echoBytes(string(bytes)))
+}
+
+self.body(func(_inputs) {
+	// Largest case capped at 2 MiB. Two backend limits stack here:
+	//   - resource 'data' field max = 3 MiB (Platforma value resource cap);
+	//   - gRPC default MaxRecvMsgSize = 4 MiB.
+	// JSON-string encoding adds 2 bytes of quotes on top of the payload,
+	// so even a 3 MiB payload trips the resource-data cap. 2 MiB still
```
DenKoren: So, the actual size limit is 3MiB - 2 bytes? Or it is smaller because of additional 'cost of transformation'?

If it is 3MiB - 2 bytes, can we then put a test at the level, so we see a break if we reduce the limit.
Because otherwise we may have green test here and broken real block that tries to pass more than 2MiB but still below 'old' limit.

tests/workflow-tengo/src/wasm/memory-cap.test.ts
```diff
@@ -0,0 +1,45 @@
+import { BackendCapability } from "@milaboratories/pl-middle-layer";
+import { tplTest } from "@platforma-sdk/test";
+
+// Guest memory cap enforcement. Each probe runs in its own sibling
+// instance (so a trap on one doesn't poison the next), and consumes a
+// targeted amount of heap. The backend's FallbackWasmMemoryBytes is 32 MiB
```
DenKoren: 'Fallback' is bad idea for expressing default. Looks like you need to fail first and then you get this.
Also, it is not good to keep here constant names that are actually defined in go code in another repo.
This will become obsolete very fast.

Keeping the contract (min, max, default) is good here. If we change this contract - we will see red test here and think of adjusting test with the comment.

tests/workflow-tengo/src/wasm/memory-cap.tpl.tengo
```diff
@@ -0,0 +1,41 @@
...
+// FallbackWasmMemoryBytes = 32 MiB when blocks declare 0. The fixture's
+// consumeMemory(bytes) allocates and writes that many bytes, which forces
+// the wasm linear memory to grow. If the request exceeds the Store's
+// memory cap, the guest traps.
+//
+// Test points:
+//   - 8 MiB → comfortably under 32 MiB → succeeds
+//   - 24 MiB → still under → succeeds
+//   - 40 MiB → over 32 MiB → traps
+//   - 64 MiB → well over → traps
```
DenKoren: This is strange. I see it named 'wayOver/well over' but it is actually exactly on the border of limit defined on backend side (and covered by comment to the test calling this template).
Is this a mistake? Reading this, and given the comment to the test in ts file, I expect it to succeed for all 4 tests. I'm totally surprised it fails for 40MiB.

Looks like either AI trapped you, or it is not clear for me how limits work and applied :(.

tests/workflow-tengo/src/wasm/memory-cap.test.ts
blackcat: I improved naming a little. And added memory limits to the SDK doc

tests/workflow-tengo/src/wasm/memory-cap.tpl.tengo
blackcat: naming imporved

tests/workflow-tengo/src/wasm/echo-roundtrip.tpl.tengo
blackcat: Added 2 tests. Now it's obvious

sdk/workflow-tengo/src/pframes-rs.lib.tengo
```diff
@@ -25,47 +25,28 @@
...
-
-// _load returns the wasm component handle, or undefined if the runtime
-// is not yet wired. Once the planned SDK module ships, replace the body
-// with an import of the wasm SDK lib and a call to its load(WASM_NAME,
-// limit) entry point.
+// `limit` is reserved for a future per-instance memory cap override; today
+// the host applies WasmV1Data.DefaultMemoryLimit stamped at install time.
 _load := func(limit) {
-	return undefined
+	wasm := assets.importWasm("@milaboratories/pframes-rs-wasip2:main")
```
vadimpiven: how this works? what is main?

sdk/workflow-tengo/src/pframes-rs.lib.tengo
```diff
@@ -25,47 +25,28 @@
...
+	if is_error(wasm) {
+		return wasm
+	}
+	return wasm["milaboratories:pframes/spec"]
```
vadimpiven: and what does this path mean?

sdk/workflow-tengo/src/pframes-rs.lib.tengo
```diff
@@ -87,16 +68,16 @@ _resolve := func(handle, path) {
 _staticCall := func(path) {
```
vadimpiven: are you sure this and other relevant methods should be here and not in some `wasm-utils.lib.tengo` which can be reused by all wasm modules?

tests/workflow-tengo/src/wasm/pframes-rs.wrapper.test.ts
```diff
@@ -0,0 +1,111 @@
...
+    if (!pl.hasCapability("wasm:v1")) {
+      skip();
+      return;
+    }
+
+    const result = await helper.renderTemplate(
+      false,
+      "wasm.pframes-rs.wrapper",
+      [
+        "buildQueryOk",
```
vadimpiven: only 3 Ok scenarios, everything else is bad/error, is that as intended? you can find positive scenarios in pframes repo if you need them, or I can guide you though the repo

tests/workflow-tengo/src/wasm/pframes-rs.wrapper.tpl.tengo
```diff
@@ -0,0 +1,135 @@
...
+	ac := pframesRs.axes.collapse(badRequest)
+	axesCollapseBad := stringify(ac)
+
+	// === frame.fromJson + instance methods =========================
+	// Try a minimal spec first; HashMap<String, _> may accept it, but
+	// downstream construction may still reject. Either way we capture
+	// the outcome.
+	frame := pframesRs.frame.fromJson({})
+	frameFromJsonErr := ""
+	frameListColumns := "<skipped: from-json failed>"
```
vadimpiven: is this ai shortcut?)

tests/workflow-tengo/src/wasm/pframes-rs.tpl.tengo
```diff
@@ -0,0 +1,42 @@
...
+self.awaitState("InputsLocked")
+self.awaitState("req", "ResourceReady")
+
+self.body(func(inputs) {
+	// buildQuery accepts a JSON object and returns a result<string,string>.
+	// We don't assume a specific schema here — the wrapper just round-trips
+	// our map through JSON to the WIT call. Even if pframes-rs treats this
+	// input as invalid, the wrapper returns a Tengo *Error containing the
+	// guest's error message (still a useful end-to-end signal).
+	resp := pframesRs.frame.buildQuery(inputs.req)
```
vadimpiven: looks awesome, exactly what I needed!

tests/workflow-tengo/src/wasm/pframes-rs.wrapper.test.ts
blackcat: The purpose of this tests is to check backend behavior and ensure it is fine despite what you'll do in the blocks or SDK. 
So basically I need only  one happy path to be sure it's working overall.

sdk/workflow-tengo/src/assets/index.lib.tengo
```diff
@@ -45,6 +45,69 @@ importAsset := func(assetName) {
...
+ *     Over-cap outputs fail the render, not the wasm call itself.
+ *
+ * @param wasmName: string - full name of the wasm component entrypoint,
+ *                           including the namespace (NPM package name),
+ *                           e.g. "@milaboratories/pframes-rs-wasip2:main".
+ *
+ * @return wasm: a tengo.Map keyed by WIT interface name —
+ *               `wasm[<iface>][<resource>][<friendlyMethodName>]`
+ *               where method names are camelCased from the WIT kebab-case
+ *               (so WIT `from-json` → `fromJson`). Bare top-level funcs
```
vgpopov: nit: You expose implemntation details in comments

sdk/workflow-tengo/src/assets/index.lib.tengo
```diff
@@ -45,6 +45,69 @@ importAsset := func(assetName) {
...
+ *               live at `wasm[<friendlyFuncName>]`. Call results
+ *               that are `result<X, string>` are lifted: Ok arm is
+ *               returned, Err arm is surfaced as a Tengo `*Error`.
+ *               Guest traps also surface as Tengo `*Error`s — the
+ *               instance is poisoned after a trap and refuses further
+ *               calls.
+ *               If `plapi.loadWasm` fails (per-render cap reached, wasm
+ *               runtime disabled, etc.), the Tengo error is returned
+ *               verbatim — check with `is_error(...)` before use.
```
vgpopov: Can we throw an exception here and not manage to caller use is_error method

sdk/workflow-tengo/src/pframes-rs.lib.tengo
```diff
@@ -25,47 +25,28 @@
...
-// is not yet wired. Once the planned SDK module ships, replace the body
-// with an import of the wasm SDK lib and a call to its load(WASM_NAME,
-// limit) entry point.
+// `limit` is reserved for a future per-instance memory cap override; today
+// the host applies WasmV1Data.DefaultMemoryLimit stamped at install time.
 _load := func(limit) {
-	return undefined
+	wasm := assets.importWasm("@milaboratories/pframes-rs-wasip2:main")
+	if is_error(wasm) {
+		return wasm
```
vgpopov: it's definitely mistake, no?

tests/workflow-tengo/src/wasm/bad-json.tpl.tengo
```diff
@@ -0,0 +1,56 @@
...
+			{ name: "sample_id", type: "i32" }
+		]
+	}))
+
+	// Each case probes a bad input then a recovery call on its OWN
+	// sibling instance. Sibling sandboxes mean an unexpected trap on
+	// one case can't bleed into another — keeps diagnosis focused.
+	probe := func(badInput) {
+		w := assets.importWasm("@platforma-tests/wasm-table-fixture:main")
+		a := w["demo:tables/api"]
```
vgpopov: why is it string? Could we use reflection for more convinient API and why it's look so strange?

sdk/workflow-tengo/src/pframes-rs.lib.tengo
blackcat: Canonical WIT interface name `<wit-package>/<interface>` from world.wit. The component declares package `milaboratories:pframes; interface spec { … } → milaboratories:pframes/spec. wasm[<iface>]` indexes the per-interface map.

sdk/workflow-tengo/src/pframes-rs.lib.tengo
blackcat: ### short
It's possible to publish many wasm atrifacts with one node package. This one is main by default.

### full
https://github.com/milaboratory/platforma/blob/MILAB-6145-webassembly-tengo/tools/tengo-builder/src/compiler/main.ts#L439-L502 -- there is a detailed doc above the func.

sdk/workflow-tengo/src/pframes-rs.lib.tengo
blackcat: In simple words 
it is `spec` because you have `spec` interface in world.wit. Add `foobar` there, got `foobar` here.

sdk/workflow-tengo/src/assets/index.lib.tengo
blackcat: Intentionally. And still got questions from Vadim. Keep it. It's important to know.

sdk/workflow-tengo/src/pframes-rs.lib.tengo
blackcat: Discussed in person.
So it was about different patters:
- my pattern is to return an error and let block/SDK dev to handle the situation;
- the repo pattern - panic (which a hate, but it is what it is...)

Agreed, that I will follow repo pattern.

sdk/workflow-tengo/src/assets/index.lib.tengo
blackcat: Discussed in personal with Vitalii and Vadim. Unfortunately my idea to build "transparent layer" and just show what happened into the WASM component didn't receive any support. With this is_error==true Vadim has nothing to do except throwing ll.panic(...).

Refactoring the code to always throw an exception here. Small change in the backend and tests must be fixed.

tests/workflow-tengo/src/wasm/bad-json.tpl.tengo
blackcat: It's literally what you have in the WIT. For the one who adds new WASM it's not a meaningless couple of symbols. Check WIT here https://github.com/milaboratory/pl/blob/MILAB-6145-webassembly-tengo/fixtures/wasm-test-fixture/wit/world.wit#L1-L4

```
package demo:tables;

interface api {
  /// Mock of pframes-rs `table` resource. Only string-typed args.
  resource table {
    /// Build a table from a JSON spec. Spec example:
    ///   {"columns":[{"name":"a","type":"i32"},{"name":"b","type":"f64"}]}
    from-json: static func(spec-json: string) -> result<table, string>;
...
```

blackcat: We could do smth like:
```
wasm := assets.importWasm("@foo/boo")
asdf := wasm.demoTables.api
```
But we will have to drop symbols from the packages name. Than potentially could cause name overlap.

tests/workflow-tengo/src/wasm/pframes-rs.wrapper.tpl.tengo
blackcat: test removed

sdk/workflow-tengo/src/pframes-rs.lib.tengo
blackcat: cleaned up!

vgpopov (approved):

PR #1657, opened 2026-05-26
feat(ts-builder): dev-server sidecar for block-ui serve

dbolotin:
## Summary

Adds dev-server sidecar mechanism: when block authors run `ts-builder serve --target block-ui`, ts-builder writes `<ui>/dist/.dev-server` (atomic JSON: `schema`, `url`, `pid`) signalling the URL where the live Vite dev server is reachable. The desktop app reads this file and routes the block view to Vite instead of the `block-ui://` folder protocol, enabling HMR. File is removed on clean shutdown; the desktop app HEAD-probes the URL before using it, so a stale sidecar safely falls back to folder mode.

- `serve --target block-ui` switches from spawning Vite to using its programmatic API so we can write the sidecar with the actually-bound URL (port may differ from 5173).
- `setupSidecarLifecycle()` registers `SIGINT`/`SIGTERM`/`SIGHUP`/`exit`/`uncaughtException` cleanup hooks *before* `createServer()` to win the race against Vite's own signal handlers.
- Dev-only Vite plugin (`apply: "serve"`) injects `'wasm-unsafe-eval'` into the index.html CSP meta tag so in-page WASM (used by `SpecDriver`) loads under dev. Production loads via `block-ui://` with `bypassCSP: true` and is unchanged. A `FOLLOW-UP` block at the top of `createViteDevConfig.ts` flags the migration debt (drop bypassCSP everywhere + security review).
- `define: { "process.env": "({})" }` so browser-side packages that runtime-reference `process.env.*` (e.g. `pf-spec-driver/logging.ts`) don't crash on load.

Consumer side (desktop app) lives in milaboratory/platforma-desktop-app#<TBD> on branch `feat/dev-server-sidecar`.

## Test plan

- [x] `pnpm build` clean
- [x] End-to-end smoke with `enter-numbers-v3` block: serve → desktop loads via Vite, label edit hot-reloads
- [x] Sidecar removed on Ctrl-C / process exit
- [x] Stale sidecar (process SIGKILL'd) → desktop HEAD probe fails → folder fallback
- [x] Port collision (5173 taken by renderer) → Vite picks 5174 → sidecar carries 5174 → desktop loads correct URL

This PR introduces a dev-server sidecar mechanism for `ts-builder serve --target block-ui`: rather than spawning Vite as a child process, it uses Vite's programmatic API to discover the actual bound URL and atomically writes a `.dev-server` JSON file that the desktop app reads to enable HMR. It also adds a dev-only Vite plugin to relax the `script-src` CSP for WASM and a `process.env` shim to silence transitive Node references in the browser.

- **Sidecar lifecycle** (`dev-server-sidecar.ts`): registers SIGINT/SIGTERM/SIGHUP/`exit`/`uncaughtException` hooks before `vite.createServer()` to win the race against Vite's own signal handlers; atomic rename prevents half-written reads; `server.close()` is monkey-patched to cover programmatic/HMR restart paths.
- **CSP plugin** (`createViteDevConfig.ts`): `apply: \"serve\"` scoped plugin rewrites the `index.html` CSP meta tag to add `'wasm-unsafe-eval'`; accompanied by a detailed follow-up block listing migration debt.
- **`process.env` define**: not guarded by `isServe`, which may affect production `vite build` runs; the CSP regex also requires a specific attribute order that could silently miss non-canonical `<meta>` tags.
</details>

Mostly safe to merge for dev-only use, but the unconditional process.env define warrants a check before landing.

The core sidecar mechanism is well-designed and the signal-ordering rationale is sound. The main concern is the process.env define being applied without an isServe guard — if createViteDevConfig is used for production builds, process.env.NODE_ENV will be undefined in bundles, potentially disabling dead-code elimination and leaving Vue dev warnings active in shipped builds. The CSP regex silently failing on non-canonical attribute order is a secondary concern.

tools/ts-builder/src/configs/utils/createViteDevConfig.ts — the unconditional process.env define and attribute-order-sensitive CSP regex both live here.
</details>

| Filename | Overview |
|----------|----------|
| tools/ts-builder/src/configs/utils/createViteDevConfig.ts | Adds CSP-relaxation plugin and process.env shim; process.env define is not guarded by isServe, potentially affecting production builds |
| tools/ts-builder/src/commands/utils/dev-server-sidecar.ts | New sidecar lifecycle module — signal cleanup logic and exit codes need attention |
| tools/ts-builder/src/commands/serve.ts | Block-UI serve path refactored to programmatic Vite API with sidecar publishing; generally sound |
| tools/ts-builder/src/commands/utils/index.ts | Re-exports new dev-server-sidecar module; trivial barrel change |

</details>

</details>

```mermaid
sequenceDiagram
    participant Author as Block Author (CLI)
    participant TS as ts-builder serve
    participant SC as SidecarLifecycle
    participant Vite as Vite Dev Server
    participant FS as dist/.dev-server
    participant Desktop as Desktop App

    Author->>TS: ts-builder serve --target block-ui
    TS->>SC: setupSidecarLifecycle(sidecarPath)
    Note over SC: registers SIGINT/SIGTERM/SIGHUP/exit/uncaughtException
    TS->>Vite: vite.createServer(configFile, port, host)
    Vite-->>TS: server instance
    TS->>Vite: server.listen()
    Vite-->>TS: resolvedUrls.local[0]
    TS->>SC: "sidecar.publish({schema:1, url, pid})"
    SC->>FS: atomic write (tmp then rename)
    TS->>Vite: patch server.close() to call sidecar.remove()

    Desktop->>FS: read .dev-server JSON
    Desktop->>Vite: HEAD probe URL
    Vite-->>Desktop: 200 OK
    Desktop->>Vite: load block UI (HMR enabled)

    Author->>TS: Ctrl-C (SIGINT)
    TS->>SC: cleanup()
    SC->>FS: fs.unlink(.dev-server)
    TS->>TS: process.exit(0)

    Note over Desktop,FS: Stale sidecar (SIGKILL path): HEAD probe fails, fallback to block-ui:// folder protocol
```
</details>

`````markdown

---

### Issue 1 of 4
tools/ts-builder/src/configs/utils/createViteDevConfig.ts:141-155
The `"process.env": "({})"` define is applied unconditionally — including when `command === "build"` and `mode === "production"`. In a production build, Vite internally sets `"process.env.NODE_ENV"` to `JSON.stringify(mode)`, but the broader `"process.env"` substitution replaces the entire object with `({})` before more-specific keys can win, making `process.env.NODE_ENV` resolve to `undefined`. Libraries like Vue use `process.env.NODE_ENV === "production"` for dead-code elimination; leaving it `undefined` can produce larger, non-optimised production bundles with dev warnings active. Guard this define with `isServe` so it only fires during `vite dev`.

```suggestion
    define: {
      "import.meta.vitest": "undefined",
      // `vite dev` does not strip `process.env.X` references the way the
      // production rolldown build does. Transitive imports (e.g.
      // `pf-spec-driver/logging.ts`, `pf-driver/logging.ts`) hit top-level
      // `process.env.MI_LOG_PFRAMES` lookups and throw
      // `ReferenceError: process is not defined` in the browser at module
      // load, leaving the block view white.
      //
      // Inlining an empty object makes every `process.env.X` resolve to
      // `undefined`. Side effect: `process.env.NODE_ENV` is also `undefined`
      // in dev — fine for our current consumers, but flagged in the
      // top-of-file follow-up block as a thing to tighten later.
      ...(isServe ? { "process.env": "({})" } : {}),
    },
```

### Issue 2 of 4
tools/ts-builder/src/commands/utils/dev-server-sidecar.ts:57-64
The signal handlers always call `process.exit(0)`, making SIGINT/SIGTERM look like a clean success to any parent process or shell. The conventional exit codes are 130 for SIGINT, 143 for SIGTERM, and 129 for SIGHUP (128 + signal number). pnpm scripts, CI systems, and process supervisors that inspect `$?` may mis-classify the termination. The idiomatic fix is to remove the handler and re-kill the process with the original signal, letting the kernel apply the default disposition.

```suggestion
  const exitOn = (signal: NodeJS.Signals) => {
    const handler = () => {
      cleanup();
      // Re-raise the signal with the kernel's default disposition so the
      // shell / pnpm sees the conventional exit status (130 for SIGINT,
      // 143 for SIGTERM, 129 for SIGHUP).
      process.removeListener(signal, handler);
      process.kill(process.pid, signal);
    };
    process.on(signal, handler);
  };
```

### Issue 3 of 4
tools/ts-builder/src/configs/utils/createViteDevConfig.ts:99-100
The CSP regex matches only when `http-equiv` appears before `content`. HTML attributes can appear in any order; if a block's `index.html` writes `<meta content="…" http-equiv="Content-Security-Policy">`, the regex silently misses the tag and WASM fails to load in dev without any warning.

```suggestion
      // Match regardless of attribute order (http-equiv before or after content).
      const cspMetaRegex =
        /<meta\b[^>]*\bhttp-equiv="Content-Security-Policy"[^>]*\bcontent="([^"]*)"[^>]*\/?>/i;
```

### Issue 4 of 4
tools/ts-builder/src/commands/serve.ts:108-111
**Server left running without sidecar on missing URL**

When `server.resolvedUrls?.local?.[0]` is undefined, the function returns early but the Vite server remains listening. The patched `server.close` that removes the sidecar is never installed, so if a test harness calls `server.close()` directly the server shuts down but cleanup wiring is absent. Consider closing the server explicitly and exiting with an error when no URL is reported.

`````

</details>

<sub>Reviews (1): Last reviewed commit: ["fix(ts-builder): register sidecar cleanu..."](https://github.com/milaboratory/platforma/commit/b17e5bdd34a7ce139cd78270758c6c6985fc22c7) | [Re-trigger Greptile](https://app.greptile.com/api/retrigger?id=33901446)</sub>

> Greptile also left **4 inline comments** on this PR.

### DIFF
--- a/tools/ts-builder/src/commands/serve.ts
+++ b/tools/ts-builder/src/commands/serve.ts
@@ -5,12 +5,10 @@ import {
   getConfigInfo,
   getGlobalOptions,
   getValidatedConfigPath,
-  installSidecarCleanup,
-  removeSidecar,
   requireTarget,
   resolveVite,
+  setupSidecarLifecycle,
   validateTargetForBrowser,
-  writeSidecar,
 } from "./utils/index";
 
 export const serveCommand = new Command("serve")
@@ -76,6 +74,14 @@ async function runBlockUiServe(opts: BlockUiServeOptions): Promise<void> {
     process.env.USE_SOURCES = "1";
   }
 
+  // Resolve sidecar path BEFORE starting Vite so lifecycle hooks register
+  // first — see `setupSidecarLifecycle` for the ordering rationale.
+  const sidecarDir = opts.sidecarOut
+    ? path.resolve(opts.sidecarOut)
+    : path.resolve(process.cwd(), "dist");
+  const sidecarPath = path.join(sidecarDir, ".dev-server");
+  const sidecar = setupSidecarLifecycle(sidecarPath);
+
   // Vite is a peer dep of ts-builder; resolve it from the consumer package
   // so we use the same version the project's config is written against.
   let vite: typeof import("vite");
@@ -104,20 +110,15 @@ async function runBlockUiServe(opts: BlockUiServeOptions): Promise<void> {
     return;
   }
 
-  const sidecarDir = opts.sidecarOut
-    ? path.resolve(opts.sidecarOut)
-    : path.resolve(process.cwd(), "dist");
-  const sidecarPath = path.join(sidecarDir, ".dev-server");
-
-  writeSidecar(sidecarPath, { schema: 1, url, pid: process.pid });
-  installSidecarCleanup(sidecarPath);
+  sidecar.publish({ schema: 1, url, pid: process.pid });
   console.log(`Wrote dev-server sidecar: ${sidecarPath}`);
 
   // Make Vite's own close path remove the sidecar too — covers programmatic
-  // shutdowns (e.g. test harnesses) where signal-based cleanup won't fire.
+  // shutdowns (e.g. test harnesses) and HMR-driven server restarts where
+  // signal-based cleanup wouldn't fire.
   const origClose = server.close.bind(server);
   server.close = async () => {
-    removeSidecar(sidecarPath);
+    sidecar.remove();
     return origClose();
   };
 }
--- a/tools/ts-builder/src/commands/serve.ts
+++ b/tools/ts-builder/src/commands/serve.ts
@@ -1,18 +1,26 @@
 import { Command } from "commander";
+import * as path from "node:path";
 import {
   executeCommand,
   getConfigInfo,
   getGlobalOptions,
   getValidatedConfigPath,
+  installSidecarCleanup,
+  removeSidecar,
   requireTarget,
   resolveVite,
   validateTargetForBrowser,
+  writeSidecar,
 } from "./utils/index";
 
 export const serveCommand = new Command("serve")
   .description("Start development server")
   .option("-p, --port <port>", "Port number")
   .option("--host <host>", "Host address")
+  .option(
+    "--sidecar-out <path>",
+    "Directory the dev-server sidecar is written into (default: ./dist for block-ui)",
+  )
   .action(async (options, command) => {
     const globalOpts = getGlobalOptions(command);
     const target = requireTarget(globalOpts);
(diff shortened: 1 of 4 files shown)


AStaroverov:
this is not ts-builder responsibility, it should be in block-tools. ts-builder - don't know nothing about platform

PR #1662, opened 2026-05-28
MILAB-6319: canonical-encode createJsonResource + trace string for deterministic CIDs

PaulNewling:
## What this fixes

Two non-canonical JSON encodings in `workflow-tengo` produced non-deterministic resource CIDs — causing `CIDConflictError` and silent cross-project dedup misses:

1. **`createJsonResource`** encoded via `json.encode`, whose Go-map iteration order varies per render, so the same logical value yielded a different `RTYPE_JSON` CID each time. This affects every caller, including the SDK-internal params built inside `pframes.processColumn` and `xsv.importFile`.
2. **`makeTrace`** (`pframes/spec.lib.tengo`) built the `pl7.app/trace` annotation via `json.encode` of step maps. That string sits *inside* a spec resource, so its inner key order escapes the backend's outer canonicalization; a randomized order poisoned the CID of every spec consumer.

Both now use **`canonical.encode`** (recursively key-sorted). The decoded values are unchanged; only the byte order becomes deterministic, so CIDs are stable across renders.

## How it was tested

- **SDK regression tests** (`canonical.test.tengo`, `spec.test.tengo`): assert `json.encode` is non-deterministic across repeated calls while `canonical.encode` is stable and key-sorted (flat and nested), and that the trace value-string is deterministic. The full `workflow-tengo` suite passes.
- **Downstream block:** in `mixcr-amplicon-alignment`, both encodings drove `CIDConflictError` in its `wf.test.ts`; with these two fixes, the createJsonResource- and trace-driven conflicts no longer occur on a fresh backend.

## Risks and downsides

- **One-time recompute on upgrade.** These CIDs change (bytes are now sorted), so existing projects recompute the affected cached results the first time they run on a backend carrying this SDK. Inherent to fixing CID non-determinism — the old CIDs were never stable.
- **Platform-wide blast radius.** Every `createJsonResource` caller across the SDK and all blocks now gets canonical bytes — intended (it closes the trap everywhere), but a behavior change for all consumers.
- **Encoding cost.** `canonical.encode` sorts keys recursively, slightly slower than `json.encode`; negligible for the small params/trace maps on the hot path.

## Scope note

A separate, related CID-determinism issue — the `hash_override` + `saveStdoutStream` pattern in MiXCR-family blocks, where rate-dependent stdout poisons the dedup CID — is **not** addressed here. It needs a different fix (an SDK stdout-only-capture primitive) and is tracked separately.

This PR fixes two CID non-determinism bugs in `workflow-tengo` by replacing `json.encode` (Go map iteration order is randomized per call) with `canonical.encode` (recursively key-sorted, RFC 8785-style) in `createJsonResource` and `makeTrace`. The decoded values are unchanged; only the byte order becomes deterministic, so resource CIDs are now stable across renders.

- **`smart.lib.tengo`**: `createJsonResource` now calls `canonical.encode`, closing the trap for every `RTYPE_JSON` resource including SDK-internal params in `pframes.processColumn` and `xsv.importFile`.
- **`pframes/spec.lib.tengo`**: `makeTrace` now canonical-encodes the `pl7.app/trace` annotation string, preventing its non-deterministic inner key order from poisoning the CID of parent spec resources.
- **Tests** (`canonical.test.tengo`, `spec.test.tengo`, `smart.test.tengo`): regression tests confirm `json.encode` is non-deterministic across 200 calls while `canonical.encode` produces stable, lexicographically sorted output for both flat and nested maps.
</details>

The two targeted fixes are correct and well-tested; the main caveat is a known one-time CID recompute on upgrade, which is unavoidable and documented in the changeset.

The changes are minimal and backed by thorough regression tests pinning exact sorted-key output. Several other `createValueResource` call sites in `workflow/bobject.lib.tengo`, `exec/runcmd.lib.tengo`, and `pframes-rs.lib.tengo` still use `json.encode` directly, so the same class of non-determinism survives in those resource types and warrants a follow-up pass.

The remaining `json.encode` usages for direct resource creation in `workflow/bobject.lib.tengo` and `exec/runcmd.lib.tengo` are worth a second look, since they bypass the now-fixed `createJsonResource`.
</details>

| Filename | Overview |
|----------|----------|
| sdk/workflow-tengo/src/smart.lib.tengo | Switches `createJsonResource` from `json.encode` to `canonical.encode`; fix is correct and well-scoped, though several sibling sites that call `createValueResource` directly with `json.encode` remain unfixed. |
| sdk/workflow-tengo/src/pframes/spec.lib.tengo | Switches `makeTrace` trace-string encoding from `json.encode` to `canonical.encode`; the `string()` cast around the result is now redundant (canonical.encode already returns a string for arrays) but harmless. |
| sdk/workflow-tengo/src/canonical.test.tengo | Adds regression tests confirming `json.encode` is non-deterministic and `canonical.encode` is stable across 200 calls for both flat and nested maps; good coverage. |
| sdk/workflow-tengo/src/pframes/spec.test.tengo | Adds `TestMakeTraceValueStrIsDeterministic` and `TestMakeTraceValueStrIsSortedKeys` to pin the exact sorted-key output of `makeTrace.valueStr`; solidly guards the MILAB-6319 regression. |
| sdk/workflow-tengo/src/smart.test.tengo | New test file that guards the `canonical.encode` contract for `createJsonResource` with golden-string assertions for flat and nested maps. |

</details>

</details>

```mermaid
flowchart TD
    A[createJsonResource / makeTrace called] --> B{Encode value}
    B -- before fix --> C[json.encode\nGo map iteration = random order]
    B -- after fix --> D[canonical.encode\nkeys sorted lexicographically]
    C --> E[Non-deterministic bytes]
    D --> F[Deterministic bytes]
    E --> G[Different CID per render\nCIDConflictError / dedup miss]
    F --> H[Stable CID across renders]
    H --> I[tx.createValue stores resource]
    G --> J[Bug]
    I --> K[Fixed]
```
</details>

`````markdown
Fix the following 1 code review issue. Work through them one at a time, proposing concise fixes.

---

### Issue 1 of 1
sdk/workflow-tengo/src/smart.lib.tengo:1514-1516
**Other resource-creation sites still use `json.encode` directly**

Several callers in the SDK bypass `createJsonResource` entirely and call `smart.createValueResource` (or `tx.createValue`) with a raw `json.encode(...)` output, leaving those resource types equally non-deterministic. For example, `workflow/bobject.lib.tengo:49` does `smart.createValueResource(constants.RTYPE_BOBJECT_SPEC, json.encode(spec))`, and `exec/runcmd.lib.tengo:773-776` creates `RTYPE_RUN_COMMAND_OPTIONS` and `RTYPE_RUN_COMMAND_ARGS` the same way. Any map-typed spec or options struct at those sites will produce non-stable CIDs the same way `createJsonResource` did before this fix. These are outside this PR's stated scope but worth tracking as follow-up.

`````

</details>

<sub>Reviews (1): Last reviewed commit: ["MILAB-6319: canonicalize trace string in..."](https://github.com/milaboratory/platforma/commit/dc182080b8c4caedcbac8139bb21bbce38a1513a) | [Re-trigger Greptile](https://app.greptile.com/api/retrigger?id=34498981)</sub>

### DIFF
--- a/.changeset/milab-6319-canonical-createjsonresource.md
+++ b/.changeset/milab-6319-canonical-createjsonresource.md
@@ -1,0 +1,5 @@
+---
+"@platforma-sdk/workflow-tengo": patch
+---
+
+Fix non-deterministic `RTYPE_JSON` resource CIDs: `createJsonResource` now encodes via `canonical.encode` (key-sorted) instead of `json.encode`, whose Go-map iteration order varied per render. The non-determinism propagated to structural consumers (`json/getField` inside `pframes.processColumn` and `xsv.importFile`), causing `CIDConflictError`. NOTE: this changes the CID of every `RTYPE_JSON` resource, so existing projects will see a one-time recompute on upgrade. See MILAB-6319.
--- a/.changeset/milab-6319-canonical-createjsonresource.md
+++ b/.changeset/milab-6319-canonical-createjsonresource.md
@@ -2,4 +2,9 @@
 "@platforma-sdk/workflow-tengo": patch
 ---
 
-Fix non-deterministic `RTYPE_JSON` resource CIDs: `createJsonResource` now encodes via `canonical.encode` (key-sorted) instead of `json.encode`, whose Go-map iteration order varied per render. The non-determinism propagated to structural consumers (`json/getField` inside `pframes.processColumn` and `xsv.importFile`), causing `CIDConflictError`. NOTE: this changes the CID of every `RTYPE_JSON` resource, so existing projects will see a one-time recompute on upgrade. See MILAB-6319.
+Fix non-deterministic resource CIDs caused by non-canonical JSON encoding in two paths:
+
+- `createJsonResource` now encodes via `canonical.encode` (key-sorted) instead of `json.encode`, whose Go-map iteration order varied per render. This affects every `RTYPE_JSON` resource, including the SDK-internal params built inside `pframes.processColumn` and `xsv.importFile`.
+- `makeTrace` (`pframes/spec.lib.tengo`) now canonical-encodes the `pl7.app/trace` annotation string. That string is embedded inside a spec resource, so its inner key order escapes the backend's outer canonicalization; a randomized order poisoned the CID of every spec consumer.
+
+Both produced `CIDConflictError` and silent cross-project dedup misses. NOTE: this changes the CID of affected resources, so existing projects will see a one-time recompute on upgrade. See MILAB-6319.
--- a/sdk/workflow-tengo/src/canonical.test.tengo
+++ b/sdk/workflow-tengo/src/canonical.test.tengo
@@ -99,3 +99,81 @@ TestRFC8785Example := func() {
     decoded := json.decode(encoded)
     test.isEqual(decoded, obj)
 }
+
+// MILAB-6319 regression tests — establish the baseline that motivates the
+// nested-string fix at pframes/spec.lib.tengo:278.
+//
+// json.encode iterates Go maps via `range` (encode.go:164, 182), which is
+// randomized per the Go language spec. canonical.encode sorts map keys via
+// maps.getKeys (which goes through slices.quickSortInPlace). For inputs that
+// end up embedded as STRING values inside other JSON resources — like the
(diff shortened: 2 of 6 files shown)


PaulNewling:
@greptileai

PR #1663, opened 2026-05-29
uikit: center PlAlert close button vertically

PaulNewling:
## What
`PlAlert`'s `closeable` close button was pinned at `top: 10px`. That put the 40px button's *center* at 30px — ~6px below the center of a single-line (~48px) alert, drifting further as the alert grew.

Center the button on the first line of text instead:

```scss
&__close-btn {
  position: absolute;
  top: 4px;     // 24px row centre − 20px (half the 40px button)
  right: 12px;
}
```

The first content row's center sits at 24px (12px padding + half the 24px row). Placing the 40px button's top edge at `24 − 20 = 4px` puts its center on that row.

A fixed `top` — not `top: 50%` — anchors the X to the first line. On a tall multi-line alert the X stays top-right rather than floating to the vertical middle. (An earlier revision used `top: 50%`; review feedback asked to keep the multi-line X top-right, hence the first-line anchor.)

`top: 4px` also matches the existing house pattern: `PlDialogModal` already pins the same `PlCloseModalBtn` with `top: 4px`.

## Why
The off-center X showed on every `closeable` PlAlert. Surfaced on the sequence-properties block (MILAB-6368).

## Blast radius
The change touches only `.pl-alert__close-btn`, which renders only when `closeable`. Across blocks + core there are 157 `PlAlert` usages; 2 are closeable — `sequence-properties` and `clonotype-enrichment` — and both improve. Non-closeable `PlAlert`s and the separate `PlNotificationAlert` (its own `closable`/CSS) stay untouched.

## Verification
Built `@milaboratories/uikit` and previewed live in the sequence-properties dev block. The X-center sits ~24px from the alert top regardless of height:

| Alert | X vs block center |
|---|---|
| 1 line | centered (delta 0) |
| 2 lines | 8px above middle (on line 1) |
| 8 lines | 68px above middle (on line 1) |

Changeset: `@milaboratories/uikit: patch` — auto-propagates a `@platforma-sdk/ui-vue` patch (`workspace:*` + `updateInternalDependencies: patch`), so blocks pick up the fix on their next ui-vue bump.

### DIFF
--- a/lib/ui/uikit/src/components/PlAlert/pl-alert.scss
+++ b/lib/ui/uikit/src/components/PlAlert/pl-alert.scss
@@ -70,13 +70,9 @@
 
   &__close-btn {
     position: absolute;
-    // 24px is a single-line alert's vertical centre (12px padding + half the
-    // 24px content row). Hard-coding it instead of `top: 50%` keeps the X on
-    // the first line — not the middle — once the text wraps.
-    top: 24px;
+    // Centre the 40px button on the first content row. The row's centre sits at
+    // 24px (12px padding + half the 24px row); top = 24 − 20 (half the button).
+    top: 4px;
     right: 12px;
-    // Centres the button on that line via its own height — stays centred if the
-    // button is ever resized, unlike a hard-coded `top: 4px` (= 24 − 40/2).
-    transform: translateY(-50%);
   }
 }
--- a/lib/ui/uikit/src/components/PlAlert/pl-alert.scss
+++ b/lib/ui/uikit/src/components/PlAlert/pl-alert.scss
@@ -70,7 +70,9 @@
 
   &__close-btn {
     position: absolute;
-    top: 10px;
+    // center vertically regardless of alert height (single- or multi-line)
+    top: 50%;
     right: 12px;
+    transform: translateY(-50%);
   }
 }
--- a/lib/ui/uikit/src/components/PlAlert/pl-alert.scss
+++ b/lib/ui/uikit/src/components/PlAlert/pl-alert.scss
@@ -70,8 +70,10 @@
 
   &__close-btn {
     position: absolute;
-    // center vertically regardless of alert height (single- or multi-line)
-    top: 50%;
+    // 24px is a single-line alert's vertical centre (12px padding + half the
+    // 24px content row). Hard-coding it instead of `top: 50%` keeps the X on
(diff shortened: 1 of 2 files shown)


PaulNewling:
@greptileai

PaulNewling:
@greptileai

lib/ui/uikit/src/components/PlAlert/pl-alert.scss
```diff
@@ -70,7 +70,9 @@
 
   &__close-btn {
     position: absolute;
-    top: 10px;
+    // center vertically regardless of alert height (single- or multi-line)
+    top: 50%;
     right: 12px;
+    transform: translateY(-50%);
```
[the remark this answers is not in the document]
PaulNewling: The close button is 40x40px (the 24px is just the masked icon inside it), so `top: 12px` would sit ~8px below center on a single-line alert and overflow the bottom — worse than the current 10px. `top: 50%` + `translateY(-50%)` centers at any height: verified deltaY 0 on both single- and multi-line alerts. Neither `closeable` PlAlert in the codebase uses a `label`, and vertical centering is the intended behavior here.

lib/ui/uikit/src/components/PlAlert/pl-alert.scss
```diff
@@ -70,7 +70,9 @@
 
   &__close-btn {
     position: absolute;
-    top: 10px;
+    // center vertically regardless of alert height (single- or multi-line)
+    top: 50%;
```
AStaroverov: PlAlert can be more than one line, in this case it will be wrong style

AStaroverov (dismissed):

lib/ui/uikit/src/components/PlAlert/pl-alert.scss
PaulNewling: @AStaroverov Which is why I thought using 50 percent would always work, no? It will always be half way

Tested with multi line:
Screenshot 2026-05-29 at 8 32 51 AM

PaulNewling:
Alerts after commit [cdce7e8](https://github.com/milaboratory/platforma/pull/1663/commits/cdce7e8060a9c34699bc3ecbe79bfb3023ff2967)

alert_8line
alert_1line

PoslavskySV (approved):

PR #1666, opened 2026-05-29
MILAB-6002: fix PlDataTable column visibility reset on re-run with changed filters

PaulNewling:
## Problem

`PlDataTable` stored column visibility as an absolute hidden set, which replaced the
block's default visibility once saved. When a column's default flipped between runs —
e.g. a filter/ranking column reverting to `optional` after its filter is removed — it
leaked visible (MILAB-6002).

## Fix

Persist the user's explicit show/hide **deviations** from the block default, not the
absolute set. Untouched columns follow their current default, so a flipped default now
resolves correctly. One `resolveColumnHidden` helper reconciles default + overrides for
both the model (`computeHiddenColumns`) and the grid (`makeColDef`), keeping them in
lockstep; `deriveColumnVisibility` is its save-path inverse, reading deviations back off
the live grid.

Two derivation edge cases are handled. An empty live grid — mounting before column defs
arrive, or tearing down — falls back to the prior saved state instead of wiping it. And
the deviation arrays are sorted, so a pure column reorder no longer reads as a state
change and forces a spurious reload.

State migrates **v7 → v8**, resetting visibility once (the absolute set can't be
converted without per-column defaults).

The deprecated `createPlDataTableV2` shares this grid UI, so it now reconciles the same
deviations through the shared `computeHiddenColumns` rather than reading `hiddenColIds` as an absolute
set — otherwise the ~29 blocks still on V2 would over-include columns once a user changed
visibility.

## Tests

`resolveColumnHidden`, `computeHiddenColumns`, `computeHiddenColumnsV2`,
`computeVisibilityDeviations`, and the v7→v8 migration — each with a MILAB-6002
regression case. New coverage:

- `deriveColumnVisibility`: the empty-grid guard, fresh derivation, and collapse to `undefined`.
- `computeVisibilityDeviations`: order-stability across a column reorder.
- Migration: the full v6→v7→v8 chain, a sourceId with no matching cache entry, and the fresh-object guarantee.
- `computeHiddenColumns`: a sorted column the user explicitly hid stays force-visible (preservation wins over the hide).
- A pinned test for the one known limitation — an explicit hide is absorbed when a column's default flips to `optional` and back.

## Verify

`antibody-tcr-lead-selection`: 2 filters → 8 columns; remove one, re-run → 7 (ex-filter
hidden). Previously stayed 8 (or errored and showed all).

## Notes

- `minor` on `model` (adds the public `resolveColumnHidden` export); `patch` on `ui-vue`.
- Fixes the deprecated `createPlDataTableV2` in lockstep (~29 blocks share the grid UI).
- v7→v8 drops custom show/hide once on upgrade; block defaults reapply.
- `migrateV7toV8` returns a fresh `pTableParams` rather than aliasing its input.
- Axis show/hide is no longer persisted (only columns carry a visibility default).

This PR fixes MILAB-6002 by replacing the absolute hidden-column set with a deviation model — only the user's explicit show/hide choices (relative to each column's block-defined default) are persisted. Untouched columns now follow their current default on every run, so a filter/ranking column whose default reverts to `optional` after removal correctly re-hides instead of staying visible. A `v7→v8` state migration resets all old absolute visibility state; block defaults then apply cleanly going forward.

- Adds `resolveColumnHidden` as the single reconciliation function shared by the model (`computeHiddenColumns`) and the grid (`makeColDef`), preventing the model/UI divergence that made the bug reproducible in two places simultaneously.
- Introduces `deriveColumnVisibility` / `computeVisibilityDeviations` on the UI side to derive deviations from the live grid; the empty-grid guard prevents teardown or pre-column-def state events from wiping saved preferences.
- Routes the deprecated `createPlDataTableV2` path through the shared `computeHiddenColumns` so ~29 blocks on V2 get the same deviation logic as V3 without a separate implementation.
</details>

Safe to merge. The deviation model is correctly implemented end-to-end, the v7→v8 migration is a clean reset rather than a lossy conversion, and the test suite covers the MILAB-6002 regression, the empty-grid timing edge case, and the documented limitation of explicit hides being absorbed when a default flips.

The core fix is well-contained: `resolveColumnHidden` is the single source of precedence shared by both the model and the grid, eliminating the drift that caused the original bug. The migration strategy (reset rather than attempt a lossless conversion that would require per-column defaults not available at migration time) is the correct call. The empty-grid guard in `deriveColumnVisibility` prevents a subtle timing hazard during grid teardown and initialization. No new public API contracts are broken; `makeColDef` / `calculateGridOptions` are internal to the package. The two known behavioral gaps (axis visibility not persisted, explicit hide absorbed when default flips to optional and back) are both documented and pinned in tests.

No files require special attention. The two pre-existing review threads about axis column visibility behavior remain open discussion points but do not block the core correctness of this change.
</details>

| Filename | Overview |
|----------|----------|
| sdk/model/src/components/PlDataTable/createPlDataTable/utils.ts | Adds `resolveColumnHidden` — the single canonical precedence function (forced-hidden → user-shown → user-hidden → optional default); well-documented and fully tested. |
| sdk/model/src/components/PlDataTable/createPlDataTable/createPlDataTableV3.ts | Threads `shownColIds` through to `computeHiddenColumns`; refactors from absolute-set logic to `resolveColumnHidden`-per-column. Correctly exports for unit testing. |
| sdk/model/src/components/PlDataTable/createPlDataTable/createPlDataTableV2.ts | Replaces the old absolute-set reader with `computeHiddenColumnsV2` (delegates to the shared V3 function), correctly bypassing sort/filter preservation that V2 handles separately. |
| sdk/model/src/components/PlDataTable/state-migration.ts | Adds `migrateV7toV8` that resets `columnVisibility` and `hiddenColIds`/`shownColIds` to null/undefined; returns fresh objects to avoid aliasing. The null-sourceId fast path is intentional and documented. |
| sdk/model/src/components/PlDataTable/typesV7.ts | Bumps normalized version to 8 and adds `shownColIds` as an optional field in both `PlDataTableGridStateCore.columnVisibility` and `PTableParamsV2`. The optional annotation is intentional for backward-compat with old serialized data. |
| sdk/ui-vue/src/components/PlAgDataTable/sources/column-visibility.ts | New module: `computeVisibilityDeviations` (sorted output for stable JSON comparison) and `deriveColumnVisibility` (empty-grid guard). Clean, well-tested, and correctly scoped to column specs only. |
| sdk/ui-vue/src/components/PlAgDataTable/PlAgDataTableV2.vue | Replaces `normalizeColumnVisibility` with `deriveColumnVisibility`; `makePartialState` now accepts `api` and `prev` to handle empty-grid timing. Axis column visibility is intentionally not persisted (documented limitation). |
| sdk/ui-vue/src/components/PlAgDataTable/sources/table-source-v2.ts | Adds `shownColIds` parameter to `calculateGridOptions` / `makeColDef`; changes `hiddenColIds` from array to `ReadonlySet` for O(1) lookup; uses `resolveColumnHidden` for visibility. |
| sdk/ui-vue/src/components/PlAgDataTable/sources/table-state-v2.ts | Adds `getShownColIds` mirroring `getHiddenColIds`; correctly uses optional chaining for backward compatibility with old persisted `columnVisibility` that lacks `shownColIds`. |

</details>

</details>

```mermaid
sequenceDiagram
    participant Block as Block (server)
    participant Model as createPlDataTableV3
    participant State as table-state-v2
    participant Vue as PlAgDataTableV2.vue
    participant Grid as AG Grid
    participant CV as column-visibility.ts

    Note over Block,CV: Save path (user toggles column visibility)
    Grid->>Vue: onStateUpdated(event)
    Vue->>CV: readGridColumnVisibility(api)
    CV-->>Vue: ColumnVisibilityState[]
    Vue->>CV: deriveColumnVisibility(cols, prev)
    CV->>CV: computeVisibilityDeviations(cols)
    CV-->>Vue: columnVisibility (deviations or undefined)
    Vue->>State: "gridState.value = {columnVisibility}"
    State->>Block: "pTableParams {hiddenColIds, shownColIds}"

    Note over Block,CV: Load path (block re-runs, state restored)
    Block->>Model: computeHiddenColumns(cols, sort, filters, hiddenSpecs, shownSpecs)
    Model->>Model: resolveColumnHidden per column
    Model-->>Block: hidden column set (visible table)
    Block->>Vue: gridState.value (with columnVisibility deviations)
    Vue->>Grid: "calculateGridOptions({hiddenColIds, shownColIds})"
    Grid->>Grid: makeColDef → resolveColumnHidden (same precedence)
    Grid-->>Vue: Column defs with correct hide flags
```
</details>

<sub>Reviews (5): Last reviewed commit: ["MILAB-6002: tighten changeset wording"](https://github.com/milaboratory/platforma/commit/effebc48581cd019187791214defd3173ed045a4) | [Re-trigger Greptile](https://app.greptile.com/api/retrigger?id=34687414)</sub>

### DIFF
--- a/sdk/model/src/components/PlDataTable/createPlDataTable/createPlDataTableV2.ts
+++ b/sdk/model/src/components/PlDataTable/createPlDataTable/createPlDataTableV2.ts
@@ -1,6 +1,7 @@
 import type {
   AxisId,
   PColumn,
+  PColumnSpec,
   PObjectId,
   PTableColumnId,
   PTableColumnIdAxis,
@@ -32,7 +33,8 @@ import { getMatchingLabelColumns } from "../labels";
 import { collectFilterSpecColumns } from "../../../filters/traverse";
 import { isEmpty } from "es-toolkit/compat";
 import { createPTableDefV2 } from "./createPTableDefV2";
-import { isColumnOptional } from "./utils";
+import { isColumnOptional, resolveColumnHidden } from "./utils";
+import type { Nil } from "@milaboratories/helpers";
 
 /**
  * @deprecated This function is deprecated and will be removed in future. Please migrate to createPlDataTable with v3 options for improved column discovery and display configuration. See createPlDataTableOptionsV3 for details on the new options format and migration guidance.
@@ -135,21 +137,15 @@ export function createPlDataTableV2<A, U>(
   const pframeHandle = ctx.createPFrame(fullColumns);
   if (!fullHandle || !pframeHandle) return undefined;
 
-  const hiddenColumns = new Set<PObjectId>(
-    ((): PObjectId[] => {
-      // Inner join works as a filter - all columns must be present
-      if (coreJoinType === "inner") return [];
-
-      const hiddenColIds = tableStateNormalized.pTableParams.hiddenColIds;
-      if (hiddenColIds !== null) {
-        return hiddenColIds
-          .filter((s): s is PTableColumnIdColumn => s.type === "column")
-          .map((s) => s.id);
-      }
-
-      return columns.filter((c) => isColumnOptional(c.spec)).map((c) => c.id);
-    })(),
-  );
+  const hiddenColumns =
+    // Inner join works as a filter - all columns must be present.
+    coreJoinType === "inner"
+      ? new Set<PObjectId>()
+      : computeHiddenColumnsV2(
+          columns,
+          tableStateNormalized.pTableParams.hiddenColIds,
+          tableStateNormalized.pTableParams.shownColIds,
+        );
 
   // Preserve linker columns
   columns.filter((c) => isLinkerColumn(c.spec)).forEach((c) => hiddenColumns.delete(c.id));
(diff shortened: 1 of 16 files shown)


PaulNewling:
@greptileai

PaulNewling:
@codecov

PaulNewling:
@greptileai

PaulNewling:
@greptileai

PaulNewling:
@greptileai

sdk/model/src/components/PlDataTable/createPlDataTable/createPlDataTableV2.ts
```diff
@@ -209,6 +205,26 @@ export function createPlDataTableV2<A, U>(
...
+ * {@link computeHiddenColumns} so the deprecated V2 path uses the exact same precedence as
+ * the V3 model and the grid's `makeColDef` — one implementation, no drift, and forced-hidden
+ * (`visibility: "hidden"`) columns are dropped here just as they are in V3.
+ *
+ * Sort/filter preservation is skipped (both args `null`); the caller then force-keeps
+ * sorted, filtered, linker, and core columns visible on top of this base set.
+ *
+ * Exported for unit testing.
+ */
+export function computeHiddenColumnsV2(
```
vadimpiven, line 220: why separate function? this is a one-liner, let's inline it

PaulNewling:
Can no longer reproduce the issue this PR is looking to resolve

PR #1690, opened 2026-06-09
feat(pframes): O(1) get-column-by-id in pf-driver + V5 wasm interfaces

vadimpiven:
## What

Two related changes around fetching a single PColumn spec by id.

### 1. pf-driver: O(1) get-column-by-id
`PFrameDriver.getColumnSpec(handle, columnId)` (called from the visualizations `PFrameFacade`) and the column lookup inside `getUniqueValues` previously did `pFrameSpec.listColumns().find(c => c.columnId === id)` — O(n) per call, round-tripping through WASM and rebuilding the full column list every time.

`PFrameHolder` already built a resolved `Record<string, PColumnSpec>` to hand to the WASM spec frame, then discarded it. It now also retains that as a JS-side `ReadonlyMap<PObjectId, PColumnSpec>` (`columnSpecs`), and both call sites read from it directly. No new public methods; the visualizations caller is unchanged and gets the speedup for free.

### 2. New self-contained wasm interfaces (V5 factory / V4 per-frame)
Adds `PFrameWasmAPIV5` and `PFrameWasmV4` to `pl-model-middle-layer`. `PFrameWasmV4` adds `getColumn(columnId): PColumnIdAndSpec | null` for O(1) lookup by id; both `getColumn` and `listColumns` return `PColumnIdAndSpec` (the new interfaces drop `PColumnInfo`).

These are fully self-contained — no `extends` of any earlier version, every method re-declared with full docs, cross-references pointing at the V4/V5 siblings. They supersede the V4 factory / V3 per-frame interfaces. Once PFrames implements `getColumn` and we adopt the new build, drivers switch to V4/V5 and every older `PFrameWasmV*`/`PFrameWasmAPIV*` interface is removed.

## Notes
- Type-checks pass for both packages; pf-driver tests (45) pass.
- The pf-driver lookup now returns the JS-side `resolveAnnotationParents(spec)` (the value fed into WASM) rather than WASM's round-tripped spec. Equivalent for current consumers, but worth a sanity check against a real plot.
- Changesets included for both `pf-driver` and `pl-model-middle-layer`.

This PR introduces two related improvements: an O(1) JS-side column lookup in `pf-driver` by retaining the resolved `columnSpecs` map on `PFrameHolder`, and a new generation of self-contained WASM interfaces (`PFrameWasmAPIV5` / `PFrameWasmV4`) plus a new `PFrameReadAPIV15` / `PTableV12` read interface that are designed to supersede their predecessors once the matching PFrames build ships.

- `PFrameHolder` now keeps the `Readonly<Record<PObjectId, PColumnSpec>>` map it already built for WASM, and `getColumnSpec` / `getUniqueValues` in `driver_impl.ts` read from it directly instead of calling `pFrameSpec.listColumns().find()` on every call.
- `PFrameWasmV4` adds a `getColumn(columnId)` method for O(1) per-frame lookup; both `getColumn` and `listColumns` return `PColumnIdAndSpec`, and `PColumnInfo` (which added a `hasData` boolean) is removed across the board.
- `PTableV12.export()` restructures the signature by moving `headers` into the `ops` bag as a required field, while `ops` itself remains optional — creating an asymmetry where an `AbortSignal` cannot be passed without also supplying headers.
</details>

Safe to merge — the runtime behaviour of existing callers is unchanged, and the new V4/V5 interfaces are additive with no consumers yet.

The JS-side map lookup is a straightforward optimisation with no behavioural change for current callers; the new WASM interfaces are declaration-only and not yet wired up. The only design note is the `headers`-in-optional-`ops` shape in `PTableV12.export()`, which is a future-facing interface with no current callers.

lib/model/middle-layer/src/pframe/internal_api/table.ts — the `PTableV12.export()` signature deserves a second look before callers adopt it.
</details>

| Filename | Overview |
|----------|----------|
| lib/node/pf-driver/src/pframe_pool.ts | Adds `columnSpecs` as a `Readonly<Record<PObjectId, PColumnSpec>>` retained from the map built for WASM; uses `Object.create(null)` for safe key-space and assigns the reference directly after building. |
| lib/node/pf-driver/src/driver_impl.ts | Both `getColumnSpec` and the column lookup in `getUniqueValues` now read from the JS-side `columnSpecs` map instead of calling `pFrameSpec.listColumns().find()`; `listColumns` still delegates to WASM. |
| lib/model/middle-layer/src/pframe/internal_api/api_wasm.ts | Adds self-contained `PFrameWasmAPIV5` factory and `PFrameWasmV4` per-frame interface; updates `PFrameWasmV3.listColumns` return type from `PColumnInfo[]` to `PColumnIdAndSpec[]`. |
| lib/model/middle-layer/src/pframe/internal_api/table.ts | Adds `PTableV12` with `export()` where `headers` is required inside an optional `ops` bag — makes it impossible to pass only an `AbortSignal`. |
| lib/model/middle-layer/src/pframe/internal_api/api_read.ts | Adds `PFrameReadAPIV15` self-contained read interface backed by `PTableV12`; structure mirrors V14 except for the updated table type. |

</details>

</details>

```mermaid
flowchart TD
    A["PFramePool.getByKey(handle)"] --> B["PFrameHolder"]
    B --> C["pFrameSpec: PFrameWasmV3\n(WASM spec frame)"]
    B --> D["columnSpecs: Readonly<Record<PObjectId, PColumnSpec>>\n(JS-side map, new)"]
    B --> E["pFrameDataPromise: Promise<PFrameV17>\n(data layer)"]
    F["getColumnSpec(handle, id)"] -->|"O(1) map read"| D
    G["getUniqueValues(handle, req)"] -->|"O(1) map read"| D
    H["listColumns(handle)"] -->|"still calls WASM"| C
    I["PFrameWasmAPIV5 (new)"] -->|"createPFrame()"| J["PFrameWasmV4 (new)\n+ getColumn(id): PColumnIdAndSpec | null\n+ listColumns(): PColumnIdAndSpec[]"]
    K["PFrameWasmAPIV4 (existing)"] -->|"createPFrame()"| L["PFrameWasmV3 (updated)\nlistColumns(): PColumnIdAndSpec[]"]
```
</details>

<details open><summary><h3>Comments Outside Diff (1)</h3></summary>

1. `lib/node/pf-driver/src/driver_impl.ts`, line 461-467 ([link](https://github.com/milaboratory/platforma/blob/51c4c8e0f37cc573e1d9f75ac649ebd11a31c1e2/lib/node/pf-driver/src/driver_impl.ts#L461-L467)) 

   <a href="#">P2</a> **`getColumnSpec` and `listColumns` now return specs from different sources**

   `getColumnSpec` (line 461-462) returns the JS-side `resolveAnnotationParents` result stored in `columnSpecs`, while `listColumns` (line 465-467) still calls `pFrameSpec.listColumns()` and returns WASM's round-tripped view of the same spec. If the WASM layer ever normalises, enriches, or drops fields during round-trip serialisation, a caller comparing results from both methods on the same column id will observe different `PColumnSpec` values. The PR notes this is "equivalent for current consumers", but the behavioural contract between these two public methods now implicitly depends on WASM being a no-op serialisation round-trip.

   `````markdown
   Path: lib/node/pf-driver/src/driver_impl.ts
   Line: 461-467

   Comment:
   **`getColumnSpec` and `listColumns` now return specs from different sources**

   `getColumnSpec` (line 461-462) returns the JS-side `resolveAnnotationParents` result stored in `columnSpecs`, while `listColumns` (line 465-467) still calls `pFrameSpec.listColumns()` and returns WASM's round-tripped view of the same spec. If the WASM layer ever normalises, enriches, or drops fields during round-trip serialisation, a caller comparing results from both methods on the same column id will observe different `PColumnSpec` values. The PR notes this is "equivalent for current consumers", but the behavioural contract between these two public methods now implicitly depends on WASM being a no-op serialisation round-trip.

   `````
   </details>
</details>

`````markdown
Fix the following 1 code review issue. Work through them one at a time, proposing concise fixes.

---

### Issue 1 of 1
lib/model/middle-layer/src/pframe/internal_api/table.ts:148-154
**`signal` cannot be passed without `headers`**

`headers` is required inside the optional `ops` bag, so `table.export(path, { signal: abortSignal })` is a TypeScript compile error — callers that only need cancellation support are forced to provide headers too. Every other async method on this table (`getFootprint`, `getShape`, `getData`) places `signal` at the top level of the optional ops bag and can receive it independently. Making `headers` optional within ops (or keeping it as a required positional argument like `PTableV11`) would restore that symmetry and avoid surprising callers.


`````

</details>

<sub>Reviews (2): Last reviewed commit: ["feat(pframes): add V12 PTable / V15 read..."](https://github.com/milaboratory/platforma/commit/65cdc71604223dc991802d0ff1da365af31b16b3) | [Re-trigger Greptile](https://app.greptile.com/api/retrigger?id=36396725)</sub>

### DIFF
--- a/.changeset/pf-driver-getcolumn-o1.md
+++ b/.changeset/pf-driver-getcolumn-o1.md
@@ -1,0 +1,5 @@
+---
+"@milaboratories/pf-driver": patch
+---
+
+pf-driver: retain the resolved column-spec map on the JS side in `PFrameHolder` and serve `getColumnSpec` (and the column lookup inside `getUniqueValues`) from it as an O(1) map read, instead of scanning `pFrameSpec.listColumns()` on every call.
--- a/.changeset/wasm-pframe-v5-getcolumn.md
+++ b/.changeset/wasm-pframe-v5-getcolumn.md
@@ -1,0 +1,5 @@
+---
+"@milaboratories/pl-model-middle-layer": patch
+---
+
+PFrames wasm: add self-contained `PFrameWasmAPIV5` factory and `PFrameWasmV4` per-frame interfaces. `PFrameWasmV4` adds `getColumn(columnId)` for O(1) single-column lookup by id; both `getColumn` and `listColumns` return `PColumnIdAndSpec` (the new interfaces drop `PColumnInfo`). These supersede the V4 factory / V3 per-frame interfaces and will replace them once the new PFrames build is adopted.
--- a/.changeset/wasm-pframe-v5-getcolumn.md
+++ b/.changeset/wasm-pframe-v5-getcolumn.md
@@ -2,4 +2,7 @@
 "@milaboratories/pl-model-middle-layer": patch
 ---
 
-PFrames wasm: add self-contained `PFrameWasmAPIV5` factory and `PFrameWasmV4` per-frame interfaces. `PFrameWasmV4` adds `getColumn(columnId)` for O(1) single-column lookup by id; both `getColumn` and `listColumns` return `PColumnIdAndSpec` (the new interfaces drop `PColumnInfo`). These supersede the V4 factory / V3 per-frame interfaces and will replace them once the new PFrames build is adopted.
+PFrames internal API: add self-contained Vx+1 interfaces that supersede the current ones and will replace them once the new PFrames build is adopted.
+
+- wasm: `PFrameWasmAPIV5` factory and `PFrameWasmV4` per-frame interface. `PFrameWasmV4` adds `getColumn(columnId)` for O(1) single-column lookup by id; both `getColumn` and `listColumns` return `PColumnIdAndSpec` (the new interfaces drop `PColumnInfo`).
+- node: `PFrameReadAPIV15` read interface and `PTableV12` table view. `PTableV12.export` moves `headers` into `ops` as an ordered `[columnIndex, headerName][]` list instead of a positional `Record<number, string>` map.
--- a/lib/model/common/src/drivers/pframe/spec/spec.ts
+++ b/lib/model/common/src/drivers/pframe/spec/spec.ts
@@ -730,12 +730,6 @@ export function getColumnIdAndSpec<Data>(column: PColumn<Data>): PColumnIdAndSpe
   };
 }
 
-/** Information returned by {@link PFrame.listColumns} method */
-export interface PColumnInfo extends PColumnIdAndSpec {
-  /** True if data was associated with this PColumn */
-  readonly hasData: boolean;
-}
-
 export interface AxisId {
   /** Type of the axis or column value. For an axis should not use non-key
    * types like float or double. */
--- a/lib/model/common/src/drivers/pframe/spec_driver.ts
+++ b/lib/model/common/src/drivers/pframe/spec_driver.ts
@@ -6,7 +6,6 @@ import type {
   AxesSpec,
   AxesId,
   PColumnIdAndSpec,
-  PColumnInfo,
   PColumnSpec,
   SingleAxisSelector,
   AxisQualification,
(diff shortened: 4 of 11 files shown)


erohinaelena (approved):

vadimpiven:
@greptileai

PR #1701, opened 2026-06-16
feat(pl-drivers): cache presigned download URLs

vadimpiven:
## What

Caches `GetDownloadURL` responses in `ClientDownload` so repeated reads of the same blob don't each pay a backend round-trip.

- New `DownloadUrlCache` (LRU, keyed by `SignedResourceId`) holding the full `{downloadUrl, headers}` response.
- `grpcGetDownloadUrl` now checks the cache before the gRPC/REST call and populates it on success.
- Each entry's TTL is derived from the expiry **encoded in the presigned URL**, not a guessed constant:
  - `X-Amz-Date` + `X-Amz-Expires` — S3 and the FS remote driver (AWS SigV4).
  - `X-Goog-Date` + `X-Goog-Expires` — GCS (V4 signed URLs).
  - The timestamp is UTC (`Z`/Zulu), verified against pl `util/storage/v4sign/presigner.go`. The compact `YYYYMMDDTHHMMSSZ` form is expanded to the extended ISO form before parsing (V8 cannot parse the compact form).
  - A **30s safety margin** is subtracted to cover clock skew between this host and pl-core plus in-flight time.
  - Local `storage://` URLs carry no encoded expiry and get a bounded 1h default TTL.

## Why

`GetDownloadURL` is called on every blob read; the returned URL is valid for its full TTL (1h for FS-remote/GCS by default), so re-requesting it per read is wasted work.

## Notes

- The TTL clock is captured atomically inside `DownloadUrlCache.set()` (not before the await), so the cache entry never outlives `expiry − margin`.
- The cache key omits `isInternalUse` because the only caller always passes `false`.
- No request coalescing: two concurrent misses for the same id will both call the backend (unchanged from before).

## Tests

`download_url_cache.test.ts` — 10 unit tests covering S3/FS, GCS, the UTC-independence guarantee, malformed/non-URL input, the 30s margin, and the no-expiry default.

This PR adds a `DownloadUrlCache` (LRU, keyed by `SignedResourceId`) to `ClientDownload` so that each `GetDownloadURL` backend call is only made once per presigned URL's lifetime instead of once per blob read. Cache entry TTLs are computed from the expiry embedded in the presigned URL (`X-Amz-Date`/`X-Amz-Expires` for S3 and the FS remote driver, `X-Goog-Date`/`X-Goog-Expires` for GCS), with a 30 s safety margin; `storage://` URLs fall back to a 1 h default.

- **`download_url_cache.ts`** — New module with `extractUrlExpiryMs`, `downloadUrlCacheTtlMs`, and `DownloadUrlCache`; the injected `logger` is currently unused, and a malformed presign date (however unlikely) would cause a fall-through to the 1 h default TTL instead of bypassing the cache.
- **`download.ts`** — `grpcGetDownloadUrl` now checks the cache before the gRPC/REST call and populates it on success; diff is minimal and correct.
- **`download_url_cache.test.ts`** — Tests cover `extractUrlExpiryMs` and `downloadUrlCacheTtlMs` thoroughly, but the `DownloadUrlCache` class itself (round-trip get/set, TTL enforcement, the `ttl ≤ 0` skip path) has no direct test coverage.
</details>

Safe to merge; the caching logic is correct for the happy path and the changes are well-scoped to a single client class.

The core TTL derivation and LRU integration are sound. The three findings are all quality/observability concerns: an unused logger, a misleading comment, and a theoretical over-caching scenario on malformed presign dates that the backend never produces in practice. Nothing in the changed path can cause data loss or incorrect auth.

download_url_cache.ts deserves a second look for the unused logger and the malformed-date fallback path; the test file is missing coverage of the DownloadUrlCache class itself.
</details>

| Filename | Overview |
|----------|----------|
| lib/node/pl-drivers/src/clients/download_url_cache.ts | New LRU cache for presigned download URLs with TTL derived from AWS/GCS query parameters; logger is accepted but unused, a misleading comment exists, and a malformed presign date falls back to the 1h default TTL instead of bypassing the cache. |
| lib/node/pl-drivers/src/clients/download.ts | Integrates DownloadUrlCache into grpcGetDownloadUrl; cache check and populate added correctly with minimal diff. |
| lib/node/pl-drivers/src/clients/download_url_cache.test.ts | 10 unit tests covering URL expiry extraction and TTL calculation; no tests for the DownloadUrlCache class itself (get/set/TTL enforcement), leaving that integration path untested. |
| .changeset/cache-download-urls.md | Changeset entry describing the patch; accurate and well-written. |
| lib/node/pl-drivers/package.json | Adds lru-cache as a direct dependency via the workspace catalog; no issues. |

</details>

</details>

`````markdown
Fix the following 3 code review issues. Work through them one at a time, proposing concise fixes.

---

### Issue 1 of 3
lib/node/pl-drivers/src/clients/download_url_cache.ts:108-111
**Misleading "Cache miss" comment**

`// Cache miss.` is the wrong term here — a cache miss means the key was not found. This path means the URL has already (nearly) expired and should not be stored. Consider `// URL already expired or within safety margin; skip caching.`

### Issue 2 of 3
lib/node/pl-drivers/src/clients/download_url_cache.ts:93-101
**`logger` is injected but never used**

`logger` is stored as `private readonly logger` but is not referenced anywhere in `get()` or `set()`. If it was intended for cache-hit/miss diagnostics it should be wired up; if not needed it can be dropped to avoid misleading future readers.

### Issue 3 of 3
lib/node/pl-drivers/src/clients/download_url_cache.ts:42-54
**Early `return undefined` on malformed expiry causes over-long default TTL**

When a URL has recognizable signing params (`X-Amz-Date` is present) but the date is malformed, line 49 returns `undefined` rather than `continue`-ing. `downloadUrlCacheTtlMs` then cannot distinguish "no expiry encoded" (`storage://`) from "expiry encoded but unreadable", and applies the full 1 h default TTL. If a near-expiring URL ever has a corrupt date field, the stale URL would be served for up to ~1 h, producing 403s from S3/GCS on every blob read until the entry ages out.

Consider returning a distinct sentinel (e.g. `null`) or throwing, so the caller knows a signing scheme was detected but the TTL cannot be trusted.

`````

</details>

<sub>Reviews (1): Last reviewed commit: ["feat(pl-drivers): cache presigned downlo..."](https://github.com/milaboratory/platforma/commit/b4098f0e2c6948735be2d871a3d6837d32c6a118) | [Re-trigger Greptile](https://app.greptile.com/api/retrigger?id=37900455)</sub>

### DIFF
--- a/lib/node/pl-drivers/src/clients/download_url_cache.ts
+++ b/lib/node/pl-drivers/src/clients/download_url_cache.ts
@@ -110,4 +110,8 @@ export class DownloadUrlCache {
     if (ttl <= 0) return; // Cache miss.
     this.cache.set(key, value, { ttl });
   }
+
+  delete(key: SignedResourceId): void {
+    this.cache.delete(key);
+  }
 }
--- a/lib/node/pl-drivers/src/clients/download_url_cache.ts
+++ b/lib/node/pl-drivers/src/clients/download_url_cache.ts
@@ -1,0 +1,113 @@
+import { LRUCache } from "lru-cache";
+import type { SignedResourceId } from "@milaboratories/pl-client";
+import type { MiLogger } from "@milaboratories/ts-helpers";
+import type { DownloadAPI_GetDownloadURL_Response } from "../proto-grpc/github.com/milaboratory/pl/controllers/shared/grpc/downloadapi/protocol";
+
+/**
+ * Safety margin subtracted from the encoded URL expiry. Covers clock skew
+ * between this host and pl-core (the timestamp is the server's signing clock,
+ * not ours) plus in-flight time between signing and use.
+ */
+const SAFETY_MARGIN_MS = 30_000;
+
+/**
+ * TTL applied to download URLs that carry no expiry in their query string -
+ * notably local `storage://` URLs, which are deterministic for a given resource
+ * and effectively never expire. Bounded so a changed storage projection is
+ * eventually re-resolved.
+ */
+const NO_EXPIRY_DEFAULT_TTL_MS = 60 * 60 * 1000; // 1h
+
+/** Max number of cached {url, headers} entries. Each entry is tiny. */
+const DEFAULT_MAX_ENTRIES = 4096;
+
+/**
+ * Extracts the absolute expiry time (epoch ms) encoded in a presigned download
+ * URL, or `undefined` if the URL carries no expiry.
+ *
+ * Supports AWS SigV4 (`X-Amz-Date` + `X-Amz-Expires`, used by pl's S3 and FS
+ * remote drivers) and GCS V4 (`X-Goog-Date` + `X-Goog-Expires`). Both encode
+ * the date as a compact ISO-8601 UTC timestamp `YYYYMMDDTHHMMSSZ` (trailing `Z`
+ * = Zulu/UTC; verified against pl `util/storage/v4sign/presigner.go`) and the
+ * lifetime as integer seconds.
+ */
+export function extractUrlExpiryMs(url: string): number | undefined {
+  let query: URLSearchParams;
+  try {
+    query = new URL(url).searchParams;
+  } catch {
+    return undefined;
+  }
+
+  for (const prefix of ["X-Amz", "X-Goog"]) {
+    const date = query.get(`${prefix}-Date`);
+    const expires = query.get(`${prefix}-Expires`);
+    if (date === null || expires === null) continue;
+
+    const signedAtMs = parseCompactIso8601Utc(date);
+    const expiresSec = Number(expires);
+    if (signedAtMs === undefined || !Number.isFinite(expiresSec)) return undefined;
+
+    return signedAtMs + expiresSec * 1000;
+  }
+
+  return undefined;
+}
+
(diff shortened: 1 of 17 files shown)

### CODE REVIEW

File: lib/node/pl-drivers/src/clients/download_url_cache.ts [hunk 1]
```diff
@@ -0,0 +1,118 @@
...
+/**
+ * Computes the cache TTL (ms from `nowMs`) for a download URL, with the safety
+ * margin already subtracted. Returns a non-positive number when the URL is
+ * already within the margin of expiring - the caller should then skip caching.
+ */
+export function downloadUrlCacheTtlMs(url: string, nowMs: number): number {
+  const expiry = extractUrlExpiryMs(url);
+  if (expiry === null) return 0;
+  if (expiry === undefined) return NO_EXPIRY_DEFAULT_TTL_MS;
+  return expiry - nowMs - SAFETY_MARGIN_MS;
```
kevindetry-milaboratories: please inline the nowMs arg as a Date.now() call

kevindetry-milaboratories (approved):

PR #1714, opened 2026-06-23
Facade: slim block packages + from-pack-v2 loader

dbolotin:
## Facade — slim block packages + `from-pack-v2`

Every block publishes a thin `block/` facade carrying only its type surface + a `from-pack-v2` runtime locator (`BlockPointer`), with **zero `@platforma-sdk/*` runtime deps**. This is the SDK-side change, shipped as one PR.

**What's here**
- **ts-builder** — `block-facade` target + facade tsconfig preset; workspace rolldown bump.
- **Structurer** — emits + enforces the facade end-state (4-file `block/src/` surface, slim `package.json`, sibling privacy, legacy shim removal); all `etc/blocks` refreshed.
- **`from-pack-v2` consumer side** — locator + loader branch, `loadPackDescriptionFromManifest`, a downstream `local-tgz` UI frontend (reuses the download driver's cache recycling), `published.json` + `--registry-serve-url`, `resolveToRegistry`.
- **E2** — monorepo test blocks migrated to the facade pointers; cross-block `from-pack-v2` load proven (`filter-column-test → table-test` result-pool export, green against a live backend).

**Spec:** `work/projects/testing-framework/facade/` in `milaboratory/text`.

**Status — WIP, opening early to exercise CI.**
- Known: `monetization-test` `check` fails — it's still on the old `BlockModel` (`export const model`) API, while the facade assumes the `BlockModelV3` `platforma` export the other 14 blocks use. Latent from the structurer refresh; fix (migrate the block or teach the structurer the old shape) is a follow-up.
- Still to come on this branch: pilot-block migration (E1), pre-publish smoke loop (E3).


etc/blocks/transfer-files/block/src/index.ts
```diff
@@ -0,0 +1,57 @@
...
+} as const;
+
+// Block-named aliases for readable cross-block imports in tests and
+// consumer code. Same types / same runtime value as the universal
+// names above; the aliases avoid `as`-renames at the import site.
+export type TransferFilesBlockContract = BlockContract;
+export type TransferFilesBlockOutputs  = BlockOutputs;
+export type TransferFilesBlockData     = BlockData;
+export type TransferFilesBlockHref     = BlockHref;
+export const TransferFilesBlockPointer = BlockPointer;
```
AStaroverov, line 53: do we really want so much aliases?

etc/blocks/upload-file/block/src/index.ts
```diff
@@ -0,0 +1,57 @@
...
+} as const;
+
+// Block-named aliases for readable cross-block imports in tests and
+// consumer code. Same types / same runtime value as the universal
+// names above; the aliases avoid `as`-renames at the import site.
+export type TestUploadFileBlockContract = BlockContract;
+export type TestUploadFileBlockOutputs  = BlockOutputs;
+export type TestUploadFileBlockData     = BlockData;
+export type TestUploadFileBlockHref     = BlockHref;
+export const TestUploadFileBlockPointer = BlockPointer;
```
AStaroverov, line 53: one a second, this is just same code as in another blocks? 
What the reason?

etc/blocks/transfer-files/block/src/index.ts
dbolotin: You just use them in different contexts. Href was used in some tests to just cast, can remove it. Data and Outputs needed, Contract is used extensively in test framework do deliver types a single pack... Will remove href.

etc/blocks/upload-file/block/src/index.ts
dbolotin: This code is generated, this is the point of the PR.

AStaroverov (approved):

PR #1751, opened 2026-07-10
MILAB-6611: pt: size ptabler CPU/RAM from input data via exec.formula

blackcat:
## What

Size the `pt` (ptabler / Polars) run's CPU and RAM from the input data volume by default, using `exec.formula`. Applied in `pt/workflow-run.tpl.tengo` — the single exec builder call in `pt/`.

When a block does **not** set an explicit `.cpu()` / `.mem()`, the runner now requests:

```
ram = between(2 GiB + 4 × f.size(), 2 GiB, 64 GiB)   fallback 4 GiB
cpu = between(2 + f.size() / 16 GiB, 2, 8)           fallback 2
```

An explicit `.cpu()` / `.mem()` still wins (unchanged). Both dimensions carry a `.staticFallback`, so the run also resolves on backends that cannot evaluate resource formulas (no `getBlobSize`).

## Why these numbers

`f.size()` sums the stored (parquet) blob size of all input files. The sizing is deliberately **moderate** because ptabler runs Polars in **lazy + streaming** mode (`collect_all(engine="streaming")`) with **disk spill** (`--spill-dir` → DuckDB sort + PFrame reads). Peak RAM is therefore a modest multiple of the input rather than the whole dataset; the 4× covers parquet→in-memory expansion plus a small operation working set, and spill handles the tail. CPU scales gently and caps low because Polars parallelizes internally and `pt` is only ever used on the medium/light queues.

## Impact

Affects the default resources for all `pt` runs (~50 call sites across blocks). Real-world usage is medium (48) / light (2) — no heavy/ui-tasks — so there is no regression against a large queue default; medium's ~1 GiB default becomes `≥ 2 GiB + 4×input`, and light requests are capped to the light limit anyway.

Not runtime-validated in isolation (needs a live backend, same as other resource formulas); relies on the monorepo CI here. Tengo build + full unit suite (201 pass / 0 fail) are green.

## Changes

- `sdk/workflow-tengo/src/pt/workflow-run.tpl.tengo` — formula sizing
- `sdk/workflow-tengo/src/pt/index.lib.tengo` — doc notes on `.cpu()`/`.mem()` defaults
- `.changeset/pt-resource-formulas.md` — `minor`

### DIFF
--- a/sdk/workflow-tengo/src/pt/workflow-run.tpl.tengo
+++ b/sdk/workflow-tengo/src/pt/workflow-run.tpl.tengo
@@ -41,12 +41,35 @@ self.body(func(inputs) {
         arg("--frame-dir").arg(inFramesFolder).
         arg("--spill-dir").arg(spillFolder).
         writeFile("workflow_sc.json", workflowJsonStruct);
-    if !is_undefined(cpu) {
-        ptCmd.cpu(cpu)
+    // Size the ptabler (Polars) run from the total input data volume, unless the
+    // block author set an explicit cpu/mem (an explicit request always wins).
+    // ptabler runs Polars in lazy + streaming mode with disk spill (--spill-dir),
+    // so peak RAM stays a modest multiple of the input size rather than the whole
+    // dataset. f.size() sums the stored (parquet) blob size of all input files via
+    // getBlobSize (a metadata read, no pre-exec). Both dimensions carry a
+    // .staticFallback so the run also resolves on backends that cannot evaluate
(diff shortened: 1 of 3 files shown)


vadimpiven (approved):

sdk/workflow-tengo/src/pt/workflow-run.tpl.tengo
vadimpiven, line 39: beware that by allocating more CPU cores you invoke more threads and so memory claims must be proportional to CPU allocation

PR #1757, opened 2026-07-15
fix: duplicateProject must not share in-flight production

vadimpiven:
## Problem

`duplicateProject` copies the `FieldsToDuplicate` set by reference with **no readiness gate**. When a project is duplicated — or shared to another user, which routes through the same function — while a block is still computing, the copy's `prodOutput`/`prodCtx`/`prodUiCtx`/`prodArgs` fields point at the **same** backend resources as the source. The copy is therefore tied to the source's live, in-flight computation instead of being an independent snapshot.

By-reference copy is safe for **completed** (`Ready`) production (immutable) — it only breaks for `NotReady` (in-flight) production.

## Fix

Duplication decides what production to copy by a cheap **resource-level** readiness scan (schema-agnostic, no model layer):

- **nothing in flight** (the common case) → copy all production as before;
- **something in flight, with model access** (an optional `ProjectHelper` + current schema) → drop the in-flight blocks **and their downstream closure**;
- **something in flight, without model access** (sharing, CLI, or an unmigrated schema) → drop **all** production. Conservative but always chain-consistent.

Dropped blocks are also removed from `blocksInLimbo`.

### Why the downstream closure

A finished block's `prodCtx` is built from its upstreams' `prodCtx` (`createProdCtx` throws if an upstream's is missing), and `check()` does not validate cross-block chains. So dropping a running upstream (block1) while keeping a finished downstream (block2) would leave block2 referencing an absent upstream context — the break flagged in review. Dropping block2 with it keeps the chain consistent.

### Why closure is opt-in, not always-on

Computing the closure needs the dependency graph, which lives in the model layer (`ProjectHelper` + QuickJS) and requires the **current** schema. Doing that unconditionally inside the resource-level `duplicateProject` would (a) regress the common path with a full mutator load on every duplicate and (b) **break duplication of un-migrated V3 projects** — `ProjectMutator.load` rejects non-current schema, which `duplicateProject` otherwise supports. So the precise path is entered **only** when a caller opts in with a helper and the source is current; everything else uses the safe resource-level fallback. Only the interactive duplicate paths (`duplicateProject`, `duplicateProjectToUser`) opt in; sharing and CLI keep the conservative default.

`ProjectMutator.productionBlocksToDropForCopy()` (new) computes in-flight ∪ downstream-closure from the mutator's graph.

## Scope

3 files: `mutator/project.ts` (the logic + the new mutator query), `middle_layer/middle_layer.ts` (two call sites opt into precision), and the changeset. No new exports, no CLI changes.

## Verification

- Full monorepo `pnpm check` passes (268/268: build, lint, format, typecheck) — the pre-push gate.
- Manual, in desktop: duplicating a project with a **running** block yields an independent copy; a finished block downstream of a running one is re-derived; stop/run in either project no longer affects the other; duplicating a **finished/idle** project shares results immediately (no regression).

<h3>Greptile Summary</h3>

This PR fixes a correctness bug in `duplicateProject` where production fields (`prodOutput`, `prodCtx`, `prodUiCtx`, `prodArgs`) were always copied by reference, tying the copy to the source's live backend computation when a block was still running. The fix introduces a three-path readiness-aware copy strategy with zero overhead for the common case.

- **Fast path** (no in-flight production): copies all production as before — no regression.
- **Precise path** (in-flight + `projectHelper` + current schema): loads `ProjectMutator`, calls `productionBlocksToDropForCopy()`, drops in-flight blocks plus their downstream closure.
- **Fallback path** (no model access): drops ALL blocks with any production — conservative but always chain-consistent.

**Key terms touched:** `FieldsToDuplicate` (unchanged set; prod fields now conditionally skipped via `dropProduction`), `prodOutput/prodCtx/prodUiCtx/prodArgs` (selectively dropped for not-ready blocks), `anyInFlight` (flag for any not-cleanly-Ready production), `dropProduction` (computed drop set), `isTargetReady` (schema-agnostic resource helper), `productionBlocksToDropForCopy()` (new public mutator method computing in-flight ∪ downstream closure), `blocksInLimbo` (filtered in copy's rendering state), `SchemaVersionCurrent` (guards `ProjectMutator.load` from unmigrated projects).

Safe to merge — the fix correctly isolates the copy from in-flight backend resources across all three paths, and the common case (all production ready) is unchanged.

The three-path logic is sound: the fast path is a no-op, the fallback is strictly conservative, and the precise path's downstream-closure traversal keeps the production chain consistent. The seeds filter in productionBlocksToDropForCopy correctly guards against passing non-graph nodes to traverseIds. No cross-repo API or type contract changes are introduced.

No files require special attention — the main logic in project.ts is well-commented and the middle_layer.ts changes are minimal.
</details>

| Filename | Overview |
|----------|----------|
| lib/node/pl-middle-layer/src/mutator/project.ts | Core fix: adds resource-level readiness scan and three-path production-drop logic in `duplicateProject`, plus new `productionBlocksToDropForCopy()` mutator method; logic is sound but `anyInFlight` naming is imprecise and `ProjectMutator.load` can now throw where duplication previously never failed |
| lib/node/pl-middle-layer/src/middle_layer/middle_layer.ts | Minimal change: both interactive duplicate call sites forward `this.env.projectHelper` to opt into the precise downstream-closure path; sharing and CLI callers in `sharing.ts` intentionally omit it |
| .changeset/duplicate-project-inflight-production.md | Changeset entry accurately describes the three-path fix and patch-level bump for `@milaboratories/pl-middle-layer` |

</details>

<a href="#gh-light-mode-only">

```mermaid
flowchart TD
    A[duplicateProject called] --> B[Read source resource data + KVs]
    B --> C{Collect prodOutput/prodCtx per block}
    C --> D[isTargetReady check per block]
    D --> E{anyInFlight?}
    E -- No --> F[Fast path: copy all production]
    E -- Yes --> G{projectHelper + schema === Current?}
    G -- Yes --> H[Precise path: ProjectMutator.load + downstream closure]
    G -- No --> I[Fallback: drop ALL blocks with any production]
    F --> J[Copy KVs, filter blocksInLimbo]
    H --> J
    I --> J
    J --> K[Copy FieldsToDuplicate, skip prodFields for dropped blocks]
    K --> L[Return new project ResourceRef]
```

</a>
<a href="#gh-dark-mode-only">

```mermaid
%%{init: {'theme': 'base', 'themeVariables': {"darkMode": true, "background": "#0d1117", "primaryColor": "#21262d", "primaryTextColor": "#e6edf3", "primaryBorderColor": "#8b949e", "lineColor": "#8b949e", "textColor": "#e6edf3", "edgeLabelBackground": "#161b22", "actorBkg": "#21262d", "actorBorder": "#8b949e", "actorTextColor": "#e6edf3", "actorLineColor": "#8b949e", "signalColor": "#8b949e", "signalTextColor": "#e6edf3", "noteBkgColor": "#373320", "noteBorderColor": "#d4a72c", "noteTextColor": "#f0e6c0", "labelBoxBkgColor": "#21262d", "labelBoxBorderColor": "#8b949e", "labelTextColor": "#e6edf3", "loopTextColor": "#e6edf3", "activationBkgColor": "#30363d", "activationBorderColor": "#8b949e"}}}%%
flowchart TD
    A[duplicateProject called] --> B[Read source resource data + KVs]
    B --> C{Collect prodOutput/prodCtx per block}
    C --> D[isTargetReady check per block]
    D --> E{anyInFlight?}
    E -- No --> F[Fast path: copy all production]
    E -- Yes --> G{projectHelper + schema === Current?}
    G -- Yes --> H[Precise path: ProjectMutator.load + downstream closure]
    G -- No --> I[Fallback: drop ALL blocks with any production]
    F --> J[Copy KVs, filter blocksInLimbo]
    H --> J
    I --> J
    J --> K[Copy FieldsToDuplicate, skip prodFields for dropped blocks]
    K --> L[Return new project ResourceRef]
```

</a>
</details>

<sub>Reviews (2): Last reviewed commit: ["refactor: make production-drop resource-..."](https://github.com/milaboratory/platforma/commit/b6a5ac5811ddb4a27d8141ab93a90fba5ce14d39) | [Re-trigger Greptile](https://app.greptile.com/api/retrigger?id=44450509)</sub>

**Context used:**

### DIFF
--- a/lib/node/pl-middle-layer/src/mutator/project.ts
+++ b/lib/node/pl-middle-layer/src/mutator/project.ts
@@ -18,6 +18,7 @@ import {
   isResourceRef,
   Pl,
   PlClient,
+  readyOrDuplicateOrError,
 } from "@milaboratories/pl-client";
 import {
   createRenderHeavyBlock,
@@ -533,23 +534,23 @@ export class ProjectMutator {
    * depend on them). Returns an empty set when the whole project's production is Ready.
    */
   public productionBlocksToDropForCopy(): Set<string> {
-    const notReady: string[] = [];
+    const dropSet = new Set<string>();
     for (const [blockId, info] of this.blockInfos) {
       const hasProduction =
         info.fields.prodOutput !== undefined || info.fields.prodCtx !== undefined;
       if (!hasProduction) continue;
       const ready =
         info.fields.prodOutput?.status === "Ready" && info.fields.prodCtx?.status === "Ready";
-      if (!ready) notReady.push(blockId);
+      if (!ready) dropSet.add(blockId);
     }
-    if (notReady.length === 0) return new Set<string>();
+    if (dropSet.size === 0) return dropSet;
 
-    // Always drop the not-ready blocks themselves; add the downstream closure of those present in
-    // the actual production graph (a partial block may lack prodArgs and thus not be a graph node —
-    // traverse() would throw on such a root, so we only seed graph-present blocks).
+    // Snapshot the not-ready blocks before we expand dropSet with their downstream closure. Only
+    // seed blocks present in the actual production graph (a partial block may lack prodArgs and thus
+    // not be a graph node — traverse() would throw on such a root, so we only seed graph-present
+    // blocks). The snapshot means expanding dropSet can't feed back into the traversal roots.
     const graph = this.getActualProductionGraph();
-    const dropSet = new Set<string>(notReady);
-    const seeds = notReady.filter((id) => graph.nodes.has(id));
+    const seeds = [...dropSet].filter((id) => graph.nodes.has(id));
     for (const id of graph.traverseIds("downstream", ...seeds)) dropSet.add(id);
     return dropSet;
   }
@@ -2042,16 +2043,15 @@ export async function duplicateProject(
     "prodArgs",
   ]);
 
-  // Resource-level readiness (schema-agnostic; no model layer). Ready == complete/immutable result,
-  // matching the mutator's own field-status rule: resourceReady or deduplicated, and not errored.
+  // Resource-level readiness (schema-agnostic; no model layer). Ready == complete/immutable and
+  // not errored: the canonical readyOrDuplicateOrError predicate (resourceReady or deduplicated or
+  // errored) with the error case excluded — so an errored production is dropped and recomputed in
+  // the copy. Matches the mutator's own field-status rule (see ProjectMutator.load).
   const isTargetReady = async (rid: SignedResourceId | undefined): Promise<boolean> => {
     if (rid === undefined || isNullSignedResourceId(rid)) return false;
     try {
       const r = await tx.getResourceData(rid, false);
-      return (
-        (r.resourceReady || isNotNullSignedResourceId(r.originalResourceId)) &&
-        isNullSignedResourceId(r.error)
-      );
+      return readyOrDuplicateOrError(r) && isNullSignedResourceId(r.error);
     } catch {
       // Resource gone/inaccessible: treat as not-ready so its production is dropped and
       // re-derived in the copy, rather than failing the whole duplicate.
--- a/lib/node/pl-middle-layer/src/mutator/project.ts
+++ b/lib/node/pl-middle-layer/src/mutator/project.ts
@@ -88,6 +88,16 @@ import type { BlockPackInfo } from "../model/block_pack";
 
 type FieldStatus = "NotReady" | "Ready" | "Error";
 
+/**
+ * A production core field is "settled" once its result is immutable — `Ready` or a finished
+ * `Error`. `NotReady` (still computing) is the only unsettled status; a missing field is unsettled.
+ * Settled production is safe to keep/share; only running production is dropped. The resource-level
+ * equivalent is {@link readyOrDuplicateOrError}.
+ */
+function isProductionFieldSettled(status: FieldStatus | undefined): boolean {
+  return status === "Ready" || status === "Error";
+}
+
 interface BlockFieldState {
   modCount: number;
   ref?: AnyRef;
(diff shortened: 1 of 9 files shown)


vadimpiven:
@greptileai

lib/node/pl-middle-layer/src/mutator/project.ts
```diff
@@ -521,6 +522,39 @@ export class ProjectMutator {
...
+   * depend on them). Returns an empty set when the whole project's production is Ready.
+   */
+  public productionBlocksToDropForCopy(): Set<string> {
+    const dropSet = new Set<string>();
+    for (const [blockId, info] of this.blockInfos) {
+      const hasProduction =
+        info.fields.prodOutput !== undefined || info.fields.prodCtx !== undefined;
+      if (!hasProduction) continue;
+      const ready =
+        info.fields.prodOutput?.status === "Ready" && info.fields.prodCtx?.status === "Ready";
```
dbolotin: Let's also keep errors. I mean those errors that are not "running and error", but just finished final error state. Sometimes you want to share a project with error to show that something went wrong, important use-case.

dbolotin: In that regard limbo blocks also should not create a problem and can be left as such.

lib/node/pl-middle-layer/src/mutator/project.ts
```diff
@@ -1957,6 +1991,14 @@ export async function duplicateProject(
...
   sourceRid: SignedResourceId,
   options?: { label?: string },
+  /**
+   * Optional model access used only to compute a precise "which production to drop" set when the
+   * source has in-flight production (see below). When omitted, or when the source is not on the
+   * current schema, duplication falls back to a conservative all-or-nothing drop — still correct,
+   * just less selective. Callers pass it to opt into precision (interactive duplicate); callers
+   * where conservative recompute is fine or a helper isn't handy (sharing, CLI) omit it.
+   */
+  projectHelper?: ProjectHelper,
```
dbolotin, line 2021: Do we have a real case where it is `undefined`? Maybe make it required?

lib/node/pl-middle-layer/src/mutator/project.ts
```diff
@@ -1980,6 +2022,80 @@ export async function duplicateProject(
...
+
+  // Resource-level readiness (schema-agnostic; no model layer). Ready == complete/immutable and
+  // not errored: the canonical readyOrDuplicateOrError predicate (resourceReady or deduplicated or
+  // errored) with the error case excluded — so an errored production is dropped and recomputed in
+  // the copy. Matches the mutator's own field-status rule (see ProjectMutator.load).
+  const isTargetReady = async (rid: SignedResourceId | undefined): Promise<boolean> => {
+    if (rid === undefined || isNullSignedResourceId(rid)) return false;
+    try {
+      const r = await tx.getResourceData(rid, false);
+      return readyOrDuplicateOrError(r) && isNullSignedResourceId(r.error);
```
dbolotin: For the shortcut path we seem to accept both error an limbo, so it is in agreement with what I suggested above for the granular per-block cleanup path.

dbolotin (approved):

PR #1763, opened 2026-07-22
MILAB-6303: template cache-key hashes sourceHash, not full source (~59% less worker CPU)

xnacly:


lib/node/pl-middle-layer/src/mutator/template/direct_template_loader_v3.ts
```diff
@@ -71,13 +71,15 @@ type Creator = <T>(
 ) => AnyResourceRef;
 
 const LibRenderer: Renderer<TemplateLibDataV3> = {
-  updateCacheKey(resource, hash, sources) {
+  updateCacheKey(resource, hash, _sources) {
```
AStaroverov, line 74: why we keep `_sources` in the args?

PoslavskySV (approved):
Good PR!

PR #1766, opened 2026-07-27
File dialog: select across folders, list sub-folders in one pass

PoslavskySV:
Part 1 of 3 for [folder-per-sample import](https://github.com/milaboratory/text/pull/201). Spec: milaboratory/text#201.

## Why

Sequencer output arrives as one folder per sample. Xencor raised this twice (onboarding, mid-pilot review); Synolo has the same layout and is hitting it inside a live demo; AZ/GSK raised it in April.

The framing was "add a flatten-view toggle", but the code says the bigger problem is elsewhere: **the selection lived on the row objects, and `Remote.vue` replaces them on every folder change** — so files from two folders could never be selected together, flatten or not. That's where the minutes went.

## What changed

**Selection survives navigation.** It moves to a path-keyed `Map` on `data`; row `selected` becomes derived. Descend → `⌘A` → descend → `⌘A` accumulates one selection. The header counter now reports the whole selection instead of just the current directory, and offers a `Clear`.

One latent bug fixed along the way: the directory row's `@click` had no `.stop`, so it bubbled to the container's deselect handler. Harmless when navigation discarded the selection anyway; with accumulation, entering a folder would have wiped everything.

**Depth-limited recursive listing.** `LsDriver.listFiles` takes an optional `{ depth, limit }`:

- `depth` walks that many levels breadth-first, reporting files from each and keeping the browsed level's directories so navigation still works. Directories below it are the frontier, not rows — a row for a folder whose contents aren't on screen isn't actionable.
- `limit` (default 5000) flags the result `truncated` rather than passing off a partial listing as complete. Unreadable directories are counted in `unreadableDirs` and stepped over.
- Omitting the options is the previous single-level behaviour, exactly.

The walk lives in `pl-model-common` as `collectListFiles(root, listDir, ops)` — a pure function over an injected single-directory lister. The driver and the UI mock share it, so they can't drift on what `depth` means, and **the traversal is unit-testable without a backend** (10 tests, `lib/model/common/src/drivers/ls.test.ts`).

**Dialog.** An "Include subfolders" control; nested rows labelled by their path below the browsed folder (every per-sample folder can hold an `R1.fastq.gz`); and when a folder holds nothing but folders, an offer to list what's inside them.

## Deliberately not here

A `recursive` flag on `LsAPI_List_Request`. Both backend drivers list one level (`drv_fs.go:428`, delimiter-scoped `drv_s3.go:274`), so client-side traversal avoids coupling this to a `pl` release for no user-visible gain at these sizes. Worth revisiting as an S3 cost optimisation — one un-delimited paginated scan would replace a call per directory.

## Verification

- `pnpm build` — 179/179 tasks (includes every package's `types:check`)
- `pnpm check` — 268/268
- `collectListFiles`: 10 new unit tests, no backend needed
- `lib/node/pl-drivers/src/drivers/ls.test.ts`: new depth/limit cases against the existing `ls_dir_structure_test` fixture — **these need a live backend and I did not run them locally**; CI will.
- uikit: 180 existing tests still pass

The dialog changes are not visually verified — I had no running desktop app. They're type- and lint-clean, and reuse existing components (`PlDropdown` in the existing search row) to keep layout risk low.

<h3>Greptile Summary</h3>

The PR adds persistent cross-folder file selection and depth-limited recursive listings.
- `ListFilesOps` defines optional traversal depth and entry limit settings.
- `ListFilesResult` is the listing response, now including truncation and unreadable-directory metadata.
- `collectListFiles` is the shared breadth-first traversal used by the real driver and UI mock.
- `LsDriver` is the local/remote storage listing interface, now accepting recursive-listing options.
- `ImportFileHandle` identifies a file selected for indexing or upload; handles are now retained in a path-keyed selection map across navigation.
- `FileDialogItem` is the dialog row type; its mutable selected field is removed because selection is derived from the map.
- `PlFileDialog` gains subfolder-depth controls, relative-path row labels, global selection count and clearing, and listing-status hints.

<details open><summary><h3>Confidence Score: 2/5</h3></summary>

The PR should not merge until stale listing responses, unbounded directory traversal beyond the entry cap, and nested-path search are corrected.

Overlapping storage requests can render handles from the wrong listing, the recursive limit does not bound directory I/O on broad trees, and search cannot match the path text introduced to distinguish nested files.

**Files Needing Attention:** lib/ui/uikit/src/components/PlFileDialog/Remote.vue, lib/ui/uikit/src/components/PlFileDialog/remote-helpers.ts, lib/model/common/src/drivers/ls.ts
</details>

| Filename | Overview |
|----------|----------|
| lib/model/common/src/drivers/ls.ts | Adds the public recursive-listing types and shared breadth-first walker, but deeper directories bypass the advertised traversal limit. |
| lib/node/pl-drivers/src/drivers/ls.ts | Refactors local and remote one-directory listing into callbacks consumed by the shared recursive walker. |
| lib/ui/uikit/src/components/PlFileDialog/Remote.vue | Adds persistent selection and recursive-listing UI, but overlapping requests can publish stale storage results. |
| lib/ui/uikit/src/components/PlFileDialog/remote-helpers.ts | Adds selection and traversal state, while search remains inconsistent with the new path-based row labels. |
| sdk/ui-vue/src/internal/test-helpers/utils.ts | Routes generated mock directory trees through the shared recursive-listing implementation. |

</details>

```mermaid
flowchart LR
  UI[PlFileDialog] -->|storage, path, depth| Driver[LsDriver.listFiles]
  Driver --> Walker[collectListFiles]
  Walker -->|one call per directory| Lister{Storage type}
  Lister -->|remote| LSAPI[lsapi]
  Lister -->|local| FS[fsp.opendir]
  Walker --> Result[ListFilesResult]
  Result --> Rows[Path-labelled rows]
  Rows --> Selection[Path-keyed selection Map]
  Selection --> Import[ImportedFiles payload]
```
</details>

<a href="https://app.greptile.com/ide/claude-code?prompt=%23%23%23%20Issue%201%0Alib%2Fui%2Fuikit%2Fsrc%2Fcomponents%2FPlFileDialog%2FRemote.vue%3A84-86%0A**Stale%20requests%20replace%20active%20listings**%0A%0AWhen%20the%20user%20switches%20between%20storages%20with%20the%20same%20initial%20path%20while%20a%20listing%20is%20pending%2C%20the%20old%20response%20passes%20the%20path-and-depth-only%20guard%20and%20replaces%20the%20active%20storage's%20rows.%20Its%20finalizer%20also%20clears%20the%20shared%20loader%2C%20and%20selecting%20those%20stale%20rows%20combines%20their%20file%20handles%20with%20the%20current%20storage%20handle%20during%20import.%0A%0A%23%23%23%20Issue%202%0Alib%2Fmodel%2Fcommon%2Fsrc%2Fdrivers%2Fls.ts%3A136-140%0A**Directory%20traversal%20bypasses%20entry%20limit**%0A%0AWhen%20a%20depth-3%20listing%20encounters%20a%20broad%20intermediate%20level%2C%20every%20deeper%20directory%20is%20enqueued%20before%20skipping%20the%20limit%20check.%20The%20walker%20then%20issues%20a%20remote%20list%20or%20local%20%60opendir%60%20call%20for%20every%20directory%20without%20reaching%20the%205%2C000-entry%20cap%2C%20leaving%20the%20file%20dialog%20slow%20or%20unresponsive%20on%20directory-heavy%20trees.%0A%0A%23%23%23%20Issue%203%0Alib%2Fui%2Fuikit%2Fsrc%2Fcomponents%2FPlFileDialog%2Fremote-helpers.ts%3A46-49%0A**Search%20ignores%20displayed%20nested%20paths**%0A%0AWhen%20a%20flattened%20listing%20displays%20a%20sample-folder%20component%20through%20%60rowLabel%60%2C%20search%20still%20matches%20only%20the%20file%20basename.%20Searching%20for%20the%20visible%20folder%20component%20therefore%20hides%20the%20corresponding%20rows%2C%20preventing%20users%20from%20locating%20nested%20files%20by%20the%20path%20introduced%20to%20distinguish%20duplicate%20filenames.%0A%0A---%0A%0AFor%20each%20issue%20above%2C%20determine%20whether%20it%20is%20valid%20and%20should%20be%20fixed.%20If%20so%2C%20fix%20it%20directly.&repo=milaboratory%2Fplatforma&pr=1766&platform=github"><picture><source media="(prefers-color-scheme: dark)" srcset="https://greptile-static-assets.s3.amazonaws.com/badges/FixAllInClaudeDark.svg?v=6"><source media="(prefers-color-scheme: light)" srcset="https://greptile-static-assets.s3.amazonaws.com/badges/FixAllInClaude.svg?v=6">Fix All in Claude Code</picture></a>

`````markdown
### Issue 1
lib/ui/uikit/src/components/PlFileDialog/Remote.vue:84-86
**Stale requests replace active listings**

When the user switches between storages with the same initial path while a listing is pending, the old response passes the path-and-depth-only guard and replaces the active storage's rows. Its finalizer also clears the shared loader, and selecting those stale rows combines their file handles with the current storage handle during import.

### Issue 2
lib/model/common/src/drivers/ls.ts:136-140
**Directory traversal bypasses entry limit**

When a depth-3 listing encounters a broad intermediate level, every deeper directory is enqueued before skipping the limit check. The walker then issues a remote list or local `opendir` call for every directory without reaching the 5,000-entry cap, leaving the file dialog slow or unresponsive on directory-heavy trees.

### Issue 3
lib/ui/uikit/src/components/PlFileDialog/remote-helpers.ts:46-49
**Search ignores displayed nested paths**

When a flattened listing displays a sample-folder component through `rowLabel`, search still matches only the file basename. Searching for the visible folder component therefore hides the corresponding rows, preventing users from locating nested files by the path introduced to distinguish duplicate filenames.

---

`````

</details>

<sub>Reviews (1): Last reviewed commit: ["File dialog: select across folders, list..."](https://github.com/milaboratory/platforma/commit/66510972e1f4f4b3e420a308cfc32c01ac9c5ff6) | [Re-trigger Greptile](https://app.greptile.com/api/retrigger?id=47771909)</sub>

- Knowledge Base — [pl-drivers](https://app.greptile.com/milaboratories/-/custom-context/knowledge-base/milaboratory/platforma/-/docs/pl-drivers.md)
- Knowledge Base — [UI Vue SDK and Component Library](https://app.greptile.com/milaboratories/-/custom-context/knowledge-base/milaboratory/platforma/-/docs/ui-vue-sdk.md)
</details>


PoslavskySV:
Pushed `ca6c5c5` after review questions on deselection and version skew.

## The skew question changed my model of this PR

I had assumed these dialog changes reach users when the desktop app upgrades. **They don't.** A block builds a self-contained UI bundle against its own pinned `@platforma-sdk/ui-vue` — samples-and-data pins `1.79.20`, and I confirmed its built bundle carries the dialog's own strings (`"Select storage"` is in `dist/assets/index-*.js`). Only `lsDriver` comes from the host.

So the two halves travel on different rails:

| Piece | Ships via |
|---|---|
| Selection, depth control, row labels | each block bumping `@platforma-sdk/ui-vue` |
| Whether `{depth}` is honoured | the desktop app's SDK bump + #524 |

Which makes **new dialog against old host the normal case**, and the old code would have rendered an "Include subfolders" dropdown that silently did nothing.

Fix: `ListFilesResult.depth` now always echoes the depth actually applied. A host that predates `ListFilesOps` returns no `depth`; the dialog learns that from its first listing and **hides the control** (and the look-inside hint). One field, no heuristics, no false positives — I'd first considered inferring it from "subdirs present but no nested entries", which misfires on empty subdirectories.

The reverse skew is safe by construction: an old bundled dialog passes no options, `collectListFiles` applies depth 1, behaviour is byte-identical to today.

## Deselection — and two bugs in the same family as the one I already fixed

The gestures now are: ctrl/⌘-click removes one file; clicking empty space in the list clears the folder on screen; `Clear` in the header clears everything. Background-click is deliberately folder-scoped rather than Finder's clear-everything — with accumulation, one stray click could otherwise discard a 50-folder selection.

None of that was documented, so `Shortcuts.vue` now lists all three plus the cross-folder gesture.

And two more places leaked into the container's deselect handler for want of `@click.stop`:

- **breadcrumbs** — navigating up dropped the current folder's picks. Harmless before, when navigation discarded the selection anyway; with accumulation it's a real loss.
- **the header** — merely opening the shortcuts tooltip deselected the visible folder.

## Verification

`pnpm build` 179/179, `pnpm check` 268/268, uikit 180 tests, `collectListFiles` 11 (one new, covering the echo and that clamped values report what was applied rather than what was asked). Still no visual verification — no running app here.

PoslavskySV:
Closing — implementation was premature. This feature is being specified first as a mispec corpus in docs/text; code will follow only after that spec is reviewed and approved. The branch is deleted; commits remain reachable from this closed PR if any of the analysis is wanted later.

PR #1767, opened 2026-07-28
MILAB-6648: Project template

AStaroverov:
<h3>Greptile Summary</h3>

This PR introduces versioned block kinds and project-template import/export across the model SDK, middle layer, registry, build tooling, and example blocks.

- **Block kind** — a separately versioned package defining and validating a block’s initialization-parameter contract; the PR adds `@platforma-sdk/block-kind`, kind references, manifests, publication, and registry resolution.
- **Initialization parameters** — kind-validated values used to create initial block storage; the PR threads template parameters through the storage facade into `DataCreateFn`.
- **Template parameters** — the serializable projection of live block data used during project-template export; the PR adds the required `templateParams()` model callback and reference wrapping.
- **Project template** — a versioned YAML document describing blocks, dependencies, kinds, and initialization parameters; the PR adds parsing, validation, serialization, resolution, application, and export.
- **Kind selector** — an exact, patch-floating, or minor-floating version constraint used to resolve a compatible kind implementation; the PR adds parsing and explicit semver-range conversion, including pre-1.0 behavior.
- **Kind registry overview** — registry metadata mapping kind versions to implementing block versions; the PR adds publication, reconciliation, stale-association removal, and resolution logic.
- **Block reference wrapper** — a serialized marker identifying template parameters that refer to another block or column identifier; the PR adds wrapping and textual remapping during template application.
- **BlockModelV3 data model** — the versioned persisted state and migration contract for a block; the PR makes block kinds and template projections part of model construction.
- **Block structure version 2** — the canonical block repository layout extended with a dedicated `kind` package; the PR updates structure rules, scaffolds, package validation, and example blocks.

The PR appears safe to merge.

No blocking failure remains; the previously reported stale kind association, pre-1.0 selector, and dropped initialization-parameter paths are addressed in the current code.
</details>

| Filename | Overview |
|----------|----------|
| sdk/block-kind/src/index.ts | Introduces the public block-kind definition and runtime parameter-validation API. |
| lib/model/common/src/template/project_template_v1.ts | Defines the shared versioned project-template document contract. |
| sdk/model/src/block_migrations.ts | Extends initial-data creation with optional kind parameters and correctly forwards validated template parameters. |
| sdk/model/src/block_model.ts | Integrates block kinds and template-parameter projection into BlockModelV3 construction. |
| lib/node/pl-middle-layer/src/model/template_apply.ts | Implements validation and application of resolved template entries into fresh project blocks. |
| lib/node/pl-middle-layer/src/model/template_export.ts | Exports live project blocks and their projected parameters into template-v1 documents. |
| tools/block-tools/src/v2/registry/registry.ts | Adds kind-overview reconciliation and removes obsolete associations when published block metadata changes. |
| tools/block-tools/src/v2/registry/kind_resolver.ts | Resolves kind selectors using explicit exact, patch-floating, and minor-floating ranges. |
| tools/block-tools/src/cmd/publish.ts | Adds kind-first, version-checked publication before writing the block facade. |
| tools/block-tools/src/structure/structure-definition.ts | Extends canonical block structure with the new kind package. |
| tools/ts-builder/src/configs/rolldown/block-kind.config.ts | Adds the bundler target used to build block-kind packages. |

</details>

```mermaid
flowchart LR
  Kind[Block kind package] -->|validates init params| Model[BlockModelV3]
  Model -->|templateParams projection| Export[Template export]
  Export --> YAML[template-v1 YAML]
  YAML --> Parse[Parse and validate]
  Parse --> Resolve[Resolve kind implementations]
  Resolve --> Init[Create initial block storage]
  Init --> Project[Fresh project]
  Kind --> Publish[Kind-first publication]
  Publish --> Registry[Registry kind overview]
  Registry --> Resolve
```
</details>

<sub>Reviews (3): Last reviewed commit: ["MILAB-6648: check a kind&#39;s required para..."](https://github.com/milaboratory/platforma/commit/3c6dfd9ebdfa774b4a2f486db39df2037f59e874) | [Re-trigger Greptile](https://app.greptile.com/api/retrigger?id=48017402)</sub>

</details>


AStaroverov:
@greptileai

sdk/block-kind/src/descriptor.ts
```diff
@@ -0,0 +1,71 @@
...
+   * `BlockParams` — so a schema that forgets a required field does not compile. The
+   * reverse (a schema that accepts less than the type allows) is not caught, which is
+   * the honest limit of this slot.
+   *
+   * Params reach it with references already in live `PlRef` form; on the pre-flight
+   * check that happens before any block exists, the reference ids are placeholders,
+   * so a parser must not treat a specific id as meaningful.
+   */
+  readonly parseTemplateParams: (value: unknown) => BlockParams;
+  readonly __PHANTOM_BLOCK_PARAMS__?: (p: BlockParams) => void;
```
dbolotin, line 70: Why do you need that, exactly? Not clear from the class docs.

sdk/block-kind/src/index.ts
```diff
@@ -0,0 +1,58 @@
+import type { CompiledBlockKind } from "./descriptor";
+
+export type { CompiledBlockKind, InferBlockParams } from "./descriptor";
+
+/** A kind's identity, as declared in its own `package.json`, plus optional behaviour. */
+export interface BlockKindMeta<BlockParams = unknown> {
```
dbolotin: Why do we need this one if we have `CompiledBlockKindV1` ?

sdk/block-kind/src/index.ts
```diff
@@ -0,0 +1,58 @@
...
+ * `meta` is the kind's identity — its own `{ name, version }`. Source it from
+ * the package's `package.json` rather than hand-typing literals, so the on-wire
+ * `{name}@{version}` cannot drift from what npm publishes and what the S3
+ * manifest records (all read the same `package.json`):
+ *
+ * ```ts
+ * // a kind package's src/index.ts
+ * import { name, version } from "../package.json" with { type: "json" };
+ * export const kind = defineBlockKind<Params>({ name, version });
+ * ```
```
dbolotin, line 34: Can we do without it, i.e. without manual code package.json data passing?

sdk/model/src/block_migrations.ts
```diff
@@ -94,7 +113,17 @@ type BuilderState<S> = {
...
+   * The params type is erased to `unknown` here: the builder checks the block's
+   * factory against its kind's params, but once stored, the only caller that has
+   * params is the template engine, which carries them as parsed file content and
+   * cannot know the type.
+   */
+  initialDataFn: DataCreateFn<S, unknown>;
+  /** Reference to the block kind this data model belongs to, if declared. */
+  kindRef?: BlockKindReference;
+  /** The kind's runtime params check, threaded with {@link kindRef}. */
+  parseTemplateParams?: (value: unknown) => unknown;
```
dbolotin: `parseInitializationParams`

sdk/block-kind/src/index.ts
```diff
@@ -0,0 +1,58 @@
...
+   * path is derived from this downstream via `npmNameToKindPath`; the descriptor
+   * carries no separate `organization` field.
+   */
+  name: string;
+  version: string;
+  /**
+   * Runtime check for params that did not come from a typed caller. Required — see
+   * {@link CompiledBlockKind}'s `parseTemplateParams` for what it must do and why it exists.
+   */
+  parseTemplateParams: (value: unknown) => BlockParams;
```
dbolotin: `parseInitializationParams`,  template is one way how we can initialize the block, another one will be tests, another MCP/AI.
Another thing here, let's not allow this function to alter parameters (like zod normalization), this logic should be on the block side. In other words if we do so, there will be a temptation to for example to fill in defaults of a certain parameters with a specific values here, and this will create an incorrect logic placement, any logic belongs to the block model code, and should change with block.

sdk/model/src/block_migrations.ts
```diff
@@ -115,19 +144,32 @@ type RecoverState = {
...
+  /** Kind reference seeded by the builder, threaded through the chain into init(). */
+  protected readonly kindRef?: BlockKindReference;
+  /** The kind's runtime params check, carried for `init()` to hand to the DataModel. */
+  protected readonly parseTemplateParams?: (value: unknown) => unknown;
 
   protected constructor(state: {
     versionChain: DataVersionKey[];
     steps: MigrationStep[];
     transferSteps?: TransferStep[];
+    kindRef?: BlockKindReference;
```
dbolotin: We should add generic type parameter here, for better readability.

lib/model/common/src/bmodel/block_kind_ref.ts
```diff
@@ -0,0 +1,59 @@
...
+ * org-qualified (see Q-0005) — to one place.
+ */
+export type BlockKindReference = Branded<string, "BlockKindReference">;
+
+/**
+ * Compose a {@link BlockKindReference} from a kind's `name`/`version`.
+ *
+ * The single place that decides how the reference is assembled. If global
+ * uniqueness later requires the name segment to be org-qualified, this is the
+ * one line that changes (Q-0005).
```
dbolotin: Please remove all references to the spec, so that all comments have standalone meaning.

lib/model/common/src/template/template_form.ts
```diff
@@ -0,0 +1,134 @@
...
+  if (typeof value === "object" && value !== null) {
+    return Object.fromEntries(
+      Object.entries(value).map(([k, v]) => [k, mapRefs(v, fn, mapBlockId)]),
+    );
+  }
+  return value;
+}
+
+/** {@link mapRefs} in the other direction. */
+function mapTemplateRefs(
```
dbolotin: Well, it seems dangerous. False-positive is possible. We should either pass a helper function to the initialization routine in the block that will allow to do something like `helper.resolve(templateRef)`, or at least should somehow limit where it can be present... Or define this information in the kind... Let's discuss.

lib/model/common/src/template/template_form.ts
```diff
@@ -0,0 +1,134 @@
...
+ * into a key would rewrite its nested ids as if they were plain properties.
+ */
+function mapRefs(
+  value: unknown,
+  fn: (ref: PlRef) => unknown,
+  mapBlockId: (blockId: string) => string,
+): unknown {
+  if (isPlRef(value)) return fn(value);
+  if (typeof value === "string") return remapColumnIdBlockIds(value, mapBlockId);
+  if (isColumnUniversalKey(value)) return remapColumnIdBlockIds(value, mapBlockId);
```
dbolotin: This one is also pretty unexpected. We should discuss.

lib/node/pl-middle-layer/src/model/template_ids.ts
```diff
@@ -0,0 +1,144 @@
...
+   */
+  assign: (templateLocalId: string) => string;
+
+  /**
+   * Publish a block as a reference target, once it exists.
+   *
+   * `blockId` is passed in rather than read back from the assignment because the
+   * project is the authority on the id its block actually got.
+   */
+  record: (templateLocalId: string, blockId: string) => void;
```
dbolotin: Let's keep thing simple, we don't need half-rendered templates, let's make `assign` add the record in the map.

lib/node/pl-middle-layer/src/model/template_ids.ts
```diff
@@ -0,0 +1,144 @@
...
+
+  /**
+   * Rewrite one entry's params from file form into live form: every
+   * `{ block, output }` becomes a `PlRef` pointing at the block that entry received.
+   *
+   * Call it with params that are present. A caller holding an entry that omitted the
+   * key substitutes `{}` before this point, so there is no absent case to decide here.
+   */
+  liveParams: (params: Record<string, unknown>) => TemplateParamsRewrite;
```
dbolotin: This is the place where it get's dangerous due to the possible false positive substitutions...

lib/model/common/src/template/project_template_v1.ts
```diff
@@ -0,0 +1,416 @@
...
+   */
+  readonly id: string;
+  readonly kind: BlockKindSelectorReference;
+  readonly block?: BlockPackReference;
+  readonly location?: BlockPackLocationReference;
+  /**
+   * The block's `BlockParams` instance — opaque here, typed by the kind. Wires
+   * to other entries appear inside it as {@link TemplateLocalRef}s.
+   */
+  readonly params?: Record<string, unknown>;
```
dbolotin: Why optional?

lib/model/common/src/template/project_template_v1.ts
```diff
@@ -0,0 +1,416 @@
...
+export type ProjectTemplateV1Entry = {
+  /**
+   * Template-local identifier, unique within the file. Names the entry for
+   * inter-block references; on export it is the block's project-local UUID,
+   * reused verbatim.
+   */
+  readonly id: string;
+  readonly kind: BlockKindSelectorReference;
+  readonly block?: BlockPackReference;
+  readonly location?: BlockPackLocationReference;
```
dbolotin: Please add comments here what each is, and in what combinations this thing is valid.

dbolotin: What should be at that location? Dev block, or fully built? Which block pack type it resolves to?

sdk/model/src/block_model.ts
```diff
@@ -108,18 +112,31 @@ interface BlockModelV3Config<
   Data,
   Plugins extends Record<string, PluginRecord> = {},
   Transfers extends Record<string, unknown> = {},
+  Params = unknown,
 > {
   renderingMode: BlockRenderingMode;
-  dataModel: DataModel<Data, Transfers>;
+  dataModel: DataModel<Data, unknown, Transfers>;
```
dbolotin: Should it be `dataModel<Data, Params, Transfers>` ?

sdk/model/src/block_model.ts
```diff
@@ -108,18 +112,31 @@ interface BlockModelV3Config<
...
+  derivePrerunArgs: ((data: unknown) => unknown) | undefined;
+  /** Projects block data back to this kind's params for template export. */
+  deriveTemplateParams: ((data: Data) => Params) | undefined;
+  /**
+   * The kind's runtime check for params supplied by a template. Read off the compiled
+   * kind rather than declared per block: the params contract belongs to the kind, and
+   * two blocks implementing it must not be able to disagree about what a valid params
+   * object is. Always present — a kind cannot omit it, and a block cannot omit a kind.
+   */
+  parseTemplateParams: (value: unknown) => unknown;
```
dbolotin: Do we really need it here if we already have `kind`?

dbolotin: Ok, I see it actually is a ref (string), not an object. Maybe let's store object, who knows if we will add more fields there?

etc/blocks/sum-numbers/kind/src/index.ts
```diff
@@ -0,0 +1,34 @@
...
+
+/**
+ * Init-params contract for the sum-numbers block — what a creator or a
+ * project template supplies to seed a new instance. The block's `BlockData` is a
+ * single field, `sources`, and it is author-supplied at creation: a template can
+ * pre-wire which upstream number columns this block sums. Optional, because a
+ * block may also be created without a template, in which case `init` falls back
+ * to an unset input.
+ */
+export type BlockParams = { sources?: PlRef[] };
```
dbolotin, line 14: It is not `PlRef`s, these are template refs, different thing!

etc/blocks/sum-numbers/kind/src/index.ts
```diff
@@ -0,0 +1,34 @@
...
+ *
+ * The reference shape is reused from `pl-model-common` rather than restated, so a
+ * hand-written entry is held to exactly what the rest of the system calls a reference
+ * — including the `__isRef: true` marker the block dependency tree is rebuilt from,
+ * whose absence would otherwise produce a block wired to nothing.
+ *
+ * `.strict()` rejects and names any key this contract does not declare, instead of
+ * dropping it and applying a block that looks configured and is not.
+ */
+const Params = z.object({ sources: z.array(PlRef).optional() }).strict();
```
dbolotin: It is an important questions, should they really be optional? In that regard, should we be able to create a template from incompletely defined blocks.

lib/model/common/src/template/template_form.ts
```diff
@@ -0,0 +1,134 @@
...
+import { createTemplateLocalRef, isTemplateLocalRef } from "./project_template_v1";
+
+/**
+ * A kind's `Params` as they appear inside a template file: the same shape, with
+ * every `PlRef` replaced by a {@link TemplateLocalRef}.
+ *
+ * A block's params in a live project carry `PlRef`s, whose `blockId` is a
+ * project-local UUID. A template cannot carry those, so the file form names the
+ * upstream entry instead. The mapping is a shape transform, not a
+ * relabeling: `{ __isRef: true, blockId, name }` becomes `{ block, output }`.
```
dbolotin: See the comment for `BlockParams` in sum numbers block.

lib/model/common/src/template/template_form.ts
```diff
@@ -0,0 +1,134 @@
...
+ *
+ * Nested carriers need no special case — `PrimaryRef`'s `column`/`filter` are
```
dbolotin: That's a big question.

lib/model/common/src/template/template_form.ts
```diff
@@ -0,0 +1,134 @@
...
+ * {@link TemplateLocalRef} naming the referenced block and output.
+ *
+ * The export half of the codec. Structural and kind-agnostic — this is what the
+ * reservation rule on {@link TemplateLocalRef} buys, and why a block's
+ * `templateParams` lambda returns its params in ordinary live form rather than
+ * rewriting references itself.
+ *
+ * A `PlRef`'s `blockId` is copied through as the template-local `id` because
+ * export names each block by the UUID it already has. Its
+ * `requireEnrichments` flag is dropped: enrichments are out of scope for templates
```
dbolotin: That's correct, yet we should put them back in the initialization lambda!

lib/model/common/src/template/project_template_v1.ts
dbolotin: Aha, found, `lib/node/pl-middle-layer/src/model/template_serializer.ts:71` not sure it will properly pick up both subtypes... Please check.

lib/node/pl-middle-layer/src/model/template_validate.ts
```diff
@@ -0,0 +1,113 @@
+import type { ProjectTemplateV1, TemplateReferenceProblem } from "@milaboratories/pl-model-common";
+import { findProjectTemplateV1ReferenceProblems } from "@milaboratories/pl-model-common";
+import { inferAllReferencedBlocks } from "./args";
+import type { TemplateApplyProblem } from "./template_apply";
+
+/**
+ * Everything wrong with a template document that can be known from the document alone.
```
dbolotin: ??

lib/node/pl-middle-layer/src/model/template_apply.ts
```diff
@@ -0,0 +1,170 @@
...
+ * for it, so an orchestrator cannot pick a different implementation than the one the
+ * document was validated against. There is no project-local id: ids are assigned by
+ * the implementation, which is what keeps the id map — template-local id to assigned
+ * id — in the one place that needs it for rewriting references. And there is no
+ * label: a template names no block instances, so the label comes from the block
+ * package's own metadata.
+ */
+export type AddBlockRequest = {
+  /** The entry's template-local id, as it appears in the document. */
+  readonly id: string;
```
dbolotin: We should not offload this id mapping logic onto the higher level. Because of all those PlRefs, ColumnIds and others, this logic is such a complicated thing, that making it universal is a big question, so I would just add addBlock that returns `id` and all the conversion logic is concentrated in the template rendering function.

lib/node/pl-middle-layer/src/model/template_apply.ts
```diff
@@ -0,0 +1,170 @@
...
+ * Notice what is absent. There is no block pack, kind or version: the entry is named
+ * by its template-local `id` and the implementation looks up what it already resolved
+ * for it, so an orchestrator cannot pick a different implementation than the one the
+ * document was validated against. There is no project-local id: ids are assigned by
+ * the implementation, which is what keeps the id map — template-local id to assigned
+ * id — in the one place that needs it for rewriting references. And there is no
+ * label: a template names no block instances, so the label comes from the block
+ * package's own metadata.
+ */
+export type AddBlockRequest = {
```
dbolotin: And where is kind or blockPack id?

lib/node/pl-middle-layer/src/model/template_apply.ts
```diff
@@ -0,0 +1,170 @@
...
+    });
+
+    if (!outcome.ok) {
+      return { added, problem: { entryId: entry.id, error: outcome.error } };
+    }
+
+    added.push({ templateLocalId: entry.id, blockId: outcome.blockId });
+  }
+
+  return { added };
```
dbolotin: So all the conversion logic should be here.

lib/node/pl-middle-layer/src/mutator/template_construct.ts
```diff
@@ -0,0 +1,125 @@
...
+   * A template names no block instances, so this cannot come from the file — and must
+   * not be derived from it, since an exported template names its entries by the source
+   * project's block ids. It reaches here from resolution, the only stage that talks to
+   * a registry.
+   *
+   * A block whose model derives a title shows that instead (`project_overview.ts`
+   * prefers the derived one), so for most blocks this is invisible. For the ones that
+   * derive none it is the name the user reads.
+   */
+  readonly label: string;
```
dbolotin, line 27: Let's not add it here, label is deprecated and is not used anywhere (I guess?), let's not poison this new code with that old API.

lib/node/pl-middle-layer/src/mutator/template_construct.ts
```diff
@@ -0,0 +1,125 @@
...
+  readonly projectHelper: ProjectHelper;
+  readonly entries: ReadonlyMap<string, PreparedTemplateEntry>;
+  readonly ids?: TemplateIdMap;
+}): TemplateApplyApi {
+  const { placer, projectHelper, entries } = deps;
+  const ids = deps.ids ?? createTemplateIdMap();
+
+  return {
+    addBlock: (request): AddBlockOutcome => {
+      const prepared = entries.get(request.id);
```
dbolotin: This is a bit "cheating", that is the whole point of isolating APIs, this function should receive all the information from the caller.

tools/block-tools/src/structure/rules/kind-package-json.ts
```diff
@@ -0,0 +1,117 @@
...
+      watch: "ts-builder build --target block-kind --watch",
+      build: "ts-builder build --target block-kind && block-tools build-kind-manifest",
+      check: "ts-builder check --target block-kind",
+    },
+    dependencies: {
+      "@platforma-sdk/block-kind": "sdk:",
+      // A kind ships a runtime params check, so it is not a types-only package. zod is
+      // the default the scaffolded parser is written against; an author who validates by
+      // hand can drop this, which is why it is seeded rather than asserted on refresh.
+      zod: "catalog:",
```
dbolotin: ☣️

tools/block-tools/src/structure/templates/static/kind/src/index.ts
```diff
@@ -0,0 +1,42 @@
...
+ * TODO(block-kind): mirror `BlockParams` here. Keep `.strict()`: it is what turns a
+ * misspelled key into a rejection naming that key, instead of one silently ignored while
+ * the block initializes blank.
+ *
+ * This is a second intentional sentinel. The schema has to produce `BlockParams`, so an
+ * empty one stops compiling the moment the contract declares a required field — the check
+ * cannot drift from the contract by being left behind. zod is the default; any function
+ * from `unknown` to `BlockParams` satisfies the slot, including a hand-written one.
+ */
+const Params = z.object({}).strict();
```
dbolotin: Let's get rid of zod?

tools/block-tools/src/v2/registry/registry.ts
```diff
@@ -56,6 +68,21 @@ type PackageUpdateInfo = {
...
+ * `kinds/{org}/{name}/overview.json`. Filled while the block manifests are
+ * already parsed in the main `updateRegistry` pass — no extra reads.
+ *
+ * `touched` is the set of block id strings re-read this pass (removed from the
+ * existing overview before their fresh implementer entries are re-added). In
+ * force mode it is `null`: the overview is fully rebuilt from the scan, so the
+ * existing content is discarded rather than filtered.
+ */
+type KindTouchAccumulator = {
+  touched: Set<string> | null;
```
dbolotin: `touchedImplemeters`/`touchedBlocks`?

tools/block-tools/src/v2/registry/registry.ts
```diff
@@ -169,6 +216,16 @@ export class BlockRegistryV2 {
...
+
+      // Seed every existing kind overview empty so kinds orphaned by
+      // migration/removal are rewritten (or deleted) this pass. The block scan
+      // above already re-enumerates every live kind ref (refs live inside block
+      // manifests); this LIST exists solely to reset orphans.
+      const kindPaths = await this.storage.listFiles(KindsPrefix);
+      for (const rel of kindPaths) {
+        if (!KindOverviewPathPattern.test(rel)) continue;
+        touchedKinds.set(KindsPrefix + rel, { touched: null, add: [] });
+      }
```
dbolotin, line 238: This is an expensive operation that is O(N) from the number of kinds == number of blocks, let's do it only for `mode == "force"` only. Please check that I am not missing something, if my understanding is correct it should work okay, even for deletion, if it is not due to real deletion of the block, which anyway is processed only for the `force` mode.

dbolotin: If we have to read all kinds each time we do this sync, let's rethink how we approach it.

dbolotin (commented):
So yes, the issue with the reference mapping is real (see  #1781), and overall by-shape recognition seems to be not the best approach here.

etc/blocks/sum-numbers/kind/src/index.ts
AStaroverov: I don't agree, it's internal entity `TemplateRef` and for developer it's meanless complexity

etc/blocks/sum-numbers/kind/src/index.ts
AStaroverov: I think it's okay to allow create don't fully predefined projects or even some data properties can be not required for running

lib/node/pl-middle-layer/src/model/template_apply.ts
AStaroverov: reimplemented

lib/node/pl-middle-layer/src/mutator/template_construct.ts
AStaroverov: reimplemented

tools/block-tools/src/v2/registry/registry.ts
AStaroverov: 1. It is already force-only. The block sits inside if (mode === "force") — opened at L219, closed at L239. The incremental pass never runs this LIST.

2. Kinds ≠ blocks. One overview file per kind package (kinds/{org}/{name}/overview.json), and many block versions implement one kind — so it is far fewer than blocks. Also, listFiles(MainPrefix) at L221, in the same branch, enumerates every version of every package: strictly more than this one.

3. One LIST, no reads. No getKindOverviewAt per kind — for a rebuild plan the existing content is discarded, so it is not fetched at all (that GET was dropped in 29741ccc7).

4. Deletion — your understanding is right, with one addition. A seeded version whose manifest is gone IS handled incrementally: its entry is marked replaced at L299, then the manifest read continues at L312, so nothing re-adds it. What the incremental pass cannot see is deletion with no seed — direct removal, migration, a kind rename — which is exactly what this pre-seed exists for.

5. Worth fixing anyway: 20 lines separate if (mode === "force") from here, so "force-only" is invisible while reading. Happy to extract forceAllVersionsToBeUpdated() and seedKindOverviewsForRebuild() so the gating reads in one line.

sdk/block-kind/src/index.ts
AStaroverov: fixed

sdk/model/src/block_migrations.ts
AStaroverov: I don't understand this point

lib/model/common/src/template/template_form.ts
AStaroverov: doesn't exist anymore

lib/node/pl-middle-layer/src/model/template_ids.ts
AStaroverov: done

lib/model/common/src/template/project_template_v1.ts
AStaroverov: now required

lib/node/pl-middle-layer/src/model/template_ids.ts
AStaroverov: outdated

lib/model/common/src/template/project_template_v1.ts
AStaroverov: add comments, made types more strictly

sdk/model/src/block_model.ts
AStaroverov: agree

sdk/model/src/block_model.ts
AStaroverov: I don't catch it

lib/model/common/src/template/template_form.ts
AStaroverov: aren't dropping anymore

lib/node/pl-middle-layer/src/model/template_apply.ts
AStaroverov: If I understand it's fixed

lib/node/pl-middle-layer/src/model/template_apply.ts
AStaroverov: reimplemented

lib/node/pl-middle-layer/src/mutator/template_construct.ts
AStaroverov: Unfortunately it's not true, we still have blocks that don't use title/subtitle

tools/block-tools/src/structure/templates/static/kind/src/index.ts
AStaroverov: removed

AStaroverov:
@greptileai

dbolotin (approved):

PR #1769, opened 2026-07-30
MILAB-6701: name column predicates after capability, not recipe class

AStaroverov:
## Problem

`isLeafColumn` and `isColumnLazy` answered two different questions under names that suggested a third one, so picking the right guard meant reading the implementation.

- `isLeafColumn` — "does this column need other columns to resolve?" (topology)
- `isColumnLazy` — "is this the concrete leaf class?" (implementation detail)
- neither — "can I read this column's data?" (what call sites actually want)

In `blocks/clonotype-browser`, the first real consumer, three of four `isLeafColumn` call sites answer the wrong question:

| Site | Intent | Used | Correct |
|---|---|---|---|
| `index.ts:138` → `expandByPartition` | read data | `isLeafColumn` | `hasReachableData` |
| `index.ts:232` → `expandByPartition` | read data | `isLeafColumn` | `hasReachableData` |
| `index.ts:203` `getLeafColumnData(...)` | read data | no guard at all | `hasReachableData` |
| `index.ts:366` | topology | `isLeafColumn` | correct |

`index.ts:203` is a live latent bug: it reads data off a persisted `ColumnUniversalId` of arbitrary shape. If a split column reaches it, `getLeafColumnData` walked past the axis-filtered layer and returned the leaf's **unsliced** data while `getSpec()` reported the sliced spec — the sheet is then built on the wrong axis, silently.

## Fix

Name each predicate after what the caller can do. Which recipe classes exist stays an implementation detail behind `recipe.id`.

```ts
isColumn(value)            // any recipe
isDataColumn(value)        // bare leaf — id names a real storage column
hasReachableData(recipe)   // data readable here, consistent with getSpec()
hasSingleDataColumn(recipe) // reads one data column, not several
```

**`hasReachableData`** narrows to `DataColumnRecipe`, the interface that exposes `getData()`. True for a bare leaf and a spec override over one — the only shapes whose data still matches their spec. An axis filter drops axes from the spec that the underlying data still carries, so the slice exists only engine-side.

**`hasSingleDataColumn`** is `getReferencedIds().length === 1`. Such a column is co-indexed by its own axes, so its spec is the whole truth about them; one that reads several only lines up after they are joined in and drags in axes its spec never mentions. That is the primary/secondary split in `createPlDataTableV3`.

This replaces a walk over recipe classes, and is both simpler and more correct: a discovered hit with an empty linker path is an ordinary column (`buildSpecQuery` emits no `linkerJoin` for it), and the class walk rejected it purely for being a `ColumnDiscoveredRecipe`. It also no longer depends on the class registry, so foreign `ColumnRecipe` implementations are handled.

**No throwing path behind the guard.** `ColumnOverriddenRecipe.wrap` picks a data-bearing variant when the wrapped recipe carries data, so `getData()` exists exactly where it can be called. The choice is structural — it follows from the id — so the same id always yields the same class whether it came from `withSpecs` or from parsing a string, and recipes stay value-objects equal iff their ids match. The branch lives in `wrap` because that is the single funnel all six construction paths go through.

### API changes

- `ColumnLazy` → the `DataColumnRecipe` interface plus the `DataColumn` dispatcher value; `ColumnLazyImpl` → `DataColumnImpl`; `ColumnLazyId` → `DataColumnId`; `ColumnLazyData` → `ColumnData`. `column_lazy.ts` → `data_column.ts`.
- Removed: `isColumnLazy` (→ `isDataColumn`), `isColumnRecipe` (→ `isColumn`), `getLeafColumnData`, and the `ColumnLazy*` names above — gone rather than aliased.
- `isLeafColumn` stays as a deprecated alias of `hasSingleDataColumn`.

The changeset is marked `minor` by decision, though the removals are breaking.

## Verification

- `types:check` — 0 errors
- `vitest` — 455 passed, 0 failed
- `linter:check` / `formatter:check` — clean
- `turbo build --filter=...@platforma-sdk/model` — 167 tasks, includes `types:check` of every dependent package

`predicates.test.ts` pins the contested boundaries: an axis-filtered column reads a single data column yet has no reachable data; an unreadable override has no `getData` at all; merging overrides neither loses nor resurrects readability.

## Out of scope

- **`blocks/clonotype-browser`** (separate repo) — its three wrong call sites need migrating once this SDK version publishes. `index.ts:203` is the one with user-visible impact.
- **`expandByPartition` signature** stays `Column[]` rather than `DataColumnRecipe[]`. A runtime guard now prevents silently-wrong data, but the contract is not yet type-enforced; tightening it would break `clonotype-browser` at compile time.
- **`types:check` does not cover `*.test.ts`** — during this work a type error in a test file passed the compiler and surfaced only at runtime. Worth a separate fix.
- Collapsing `createPlDataTableV3` Form B into a single list, which would remove the last consumer-side need for `hasSingleDataColumn`.

<h3>Greptile Summary</h3>

This PR replaces recipe-class-oriented column terminology with capability-oriented predicates and data-access types.
- `ColumnRecipe`: immutable description of how to obtain a logical column; unchanged conceptually, but its general predicate is now `isColumn`.
- `DataColumnRecipe`: capability interface for recipes whose data can be read directly and remains consistent with `getSpec()`; newly introduced for bare data columns and readable spec overrides.
- `DataColumn`: dispatcher for constructing bare storage-column recipes; replaces the `ColumnLazy` dispatcher.
- `DataColumnImpl`: concrete implementation of a bare storage column backed by a `PObjectId`; replaces `ColumnLazyImpl`.
- `ColumnData`: directly readable column payload type; replaces `ColumnLazyData`.
- `isDataColumn`: identifies a bare storage column with a physical `PObjectId`; replaces `isColumnLazy`.
- `hasReachableData`: identifies recipes exposing safe direct data access, including readable overrides over bare columns.
- `hasSingleDataColumn`: identifies recipes depending on exactly one physical data column and is now used for primary-versus-linker topology classification.
- `isLeafColumn`: retained as a deprecated alias of `hasSingleDataColumn`.
- `ColumnOverriddenRecipe`: now selects a data-bearing implementation when wrapping a bare data column, allowing metadata/spec overrides to preserve direct data access.
- `getLeafColumnData`: removed because traversing filtered wrappers could return unsliced data inconsistent with the exposed spec.
- `expandByPartition`: now checks `hasReachableData` before reading input data instead of traversing to an underlying leaf.
- Data-table discovery and construction now classify primary columns through `hasSingleDataColumn`.
- Documentation, migration guidance, tests, providers, and public exports are updated for the renamed API.

The PR appears safe to merge.

No blocking failure remains.
</details>

| Filename | Overview |
|----------|----------|
| sdk/model/src/columns/data_column.ts | Renames the bare-column implementation and introduces the `DataColumnRecipe` direct-data capability interface and dispatcher. |
| sdk/model/src/columns/column_recipes/column_overrided_recipe.ts | Adds a data-bearing override variant selected structurally when the wrapped recipe is a bare data column. |
| sdk/model/src/columns/utils.ts | Introduces capability-oriented predicates, retains the deprecated topology alias, and removes unsafe leaf-data traversal. |
| sdk/model/src/columns/expand_by_partition.ts | Restricts direct partition-data inspection to recipes accepted by `hasReachableData`. |
| sdk/model/src/components/PlDataTable/createPlDataTable/createPlDataTableV3.ts | Replaces recipe-class topology classification with the single-data-column capability predicate. |
| sdk/model/src/columns/predicates.test.ts | Covers readable overrides, unreadable axis filters, merged overrides, and the distinction between topology and direct data access. |
| sdk/model/src/columns/index.ts | Updates the public column API exports from the removed lazy-column module to the data-column module. |

</details>

```mermaid
flowchart TD
  R[ColumnRecipe] --> S{hasSingleDataColumn}
  S -->|true| P[Eligible as direct/primary column]
  S -->|false| L[Requires linker-aware handling]
  R --> D{hasReachableData}
  D -->|true| G[DataColumnRecipe.getData]
  D -->|false| H[Materialize recipe id host-side]
  B[Bare DataColumn] --> D
  O[Override over bare DataColumn] --> D
  F[Axis-filtered recipe] --> H
  J[Joined/discovered recipe] --> H
```
</details>

<sub>Reviews (2): Last reviewed commit: ["MILAB-6701: settle the column predicate ..."](https://github.com/milaboratory/platforma/commit/0dee70b9910c5e403610ed1fc20d34f46d3fae3e) | [Re-trigger Greptile](https://app.greptile.com/api/retrigger?id=48687358)</sub>

**Context used:**

### DIFF
--- a/.changeset/column-data-access-api.md
+++ b/.changeset/column-data-access-api.md
@@ -10,19 +10,23 @@ that suggested a third one, so picking the right guard meant reading the
 implementation. Three call sites out of four in the first consumer block picked
 the wrong one.
 
-- `hasDirectData(recipe)` — new. Narrows to `DataColumn`, the interface that
+- `hasReachableData(recipe)` — new. Narrows to `DataColumn`, the interface that
   exposes `getData()`. True for a bare leaf and for a spec-override over one:
   the only shapes whose data still matches their spec.
 - `isSelfContained(recipe)` — new. Whether the column resolves without pulling
   in other columns. Replaces `isLeafColumn`, same semantics.
 - `ColumnLazy` → `DataColumn`, `ColumnLazyImpl` → `DataColumnImpl`,
-  `ColumnLazyId` / `ColumnLazyData` → `DataColumnId` / `DataColumnData`.
+  `ColumnLazyId` / `ColumnLazyData` → `DataColumnId` / `ColumnData`.
   Deprecated aliases keep existing imports compiling.
-- `ColumnOverriddenRecipe` now implements `getData()`, delegating to its inner
-  leaf — spec overrides never reshape data.
+- A spec override over a leaf is now readable too: `ColumnOverriddenRecipe.wrap`
+  picks a data-bearing variant when the wrapped recipe carries data, so
+  `getData()` exists exactly where it can be called and there is no throwing
+  path behind the guard. The choice is structural — it follows from the id — so
+  the same id always yields the same class whether it came from `withSpecs` or
+  from parsing a string.
 - `isColumnLazy` is deprecated. Its one remaining use is a legacy slot typed
   `PObjectId` (`PColumn.id`, `PColumnIdAndSpec.columnId`).
 - `getLeafColumnData` is deprecated **and behaves differently**: it used to walk
   past an axis-filtered layer and return the underlying leaf's *unsliced* data,
   which does not match the recipe's spec. It now returns `undefined` for
-  anything `hasDirectData` rejects.
+  anything `hasReachableData` rejects.
--- a/.changeset/column-data-access-api.md
+++ b/.changeset/column-data-access-api.md
@@ -1,0 +1,28 @@
+---
+"@platforma-sdk/model": minor
+---
+
+Name the column predicates after what a caller can do, not after the recipe classes
+
+`isLeafColumn` and `isColumnLazy` answered two different questions ("does this
+column need other columns to resolve?" and "can I read its data?") under names
+that suggested a third one, so picking the right guard meant reading the
+implementation. Three call sites out of four in the first consumer block picked
+the wrong one.
+
+- `hasDirectData(recipe)` — new. Narrows to `DataColumn`, the interface that
+  exposes `getData()`. True for a bare leaf and for a spec-override over one:
+  the only shapes whose data still matches their spec.
+- `isSelfContained(recipe)` — new. Whether the column resolves without pulling
+  in other columns. Replaces `isLeafColumn`, same semantics.
+- `ColumnLazy` → `DataColumn`, `ColumnLazyImpl` → `DataColumnImpl`,
+  `ColumnLazyId` / `ColumnLazyData` → `DataColumnId` / `DataColumnData`.
+  Deprecated aliases keep existing imports compiling.
+- `ColumnOverriddenRecipe` now implements `getData()`, delegating to its inner
+  leaf — spec overrides never reshape data.
+- `isColumnLazy` is deprecated. Its one remaining use is a legacy slot typed
+  `PObjectId` (`PColumn.id`, `PColumnIdAndSpec.columnId`).
+- `getLeafColumnData` is deprecated **and behaves differently**: it used to walk
+  past an axis-filtered layer and return the underlying leaf's *unsliced* data,
+  which does not match the recipe's spec. It now returns `undefined` for
+  anything `hasDirectData` rejects.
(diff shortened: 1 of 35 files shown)


AStaroverov:
@greptileai

kevindetry-milaboratories:
Let's add a general documentation of the new API (migration doc is good, but doesn't always apply). Like, the surface of it, so that neither human nor AI would have to manually scan the exports.

kevindetry-milaboratories:
Please make sure all the exports actually should be exports. Every piece of API should have a good justification to be public. I've noticed some partitions-related stuff, for example, and not sure if consumers should use those directly.

kevindetry-milaboratories (approved):

PR #1779, opened 2026-08-11
MILAB-6714: fail CI builds that never push their docker images

xnacly:
## Problem

Docker auto-push defaults to `isCI() && !isPrivate`, but the entrypoint descriptor is written with the image tag either way. A stray `"private": true` on a software package therefore produces a green CI run and a published block that 404s pulling its image at runtime, with nothing in the logs indicating anything went wrong.

This is what happened to `gpu-test` (MILAB-6714). The structurer stopped emitting `private` on software packages in `block-tools@2.11.6`, but that only helps blocks someone refreshes with the released tool. gpu-test was refreshed four weeks after that fix and still got the flag, because the refresh ran a stale local build.

## Change

`assertDockerPushIntent` in `lib/node/package-builder-lib/src/core/push-guard.ts`. In CI, building docker images without pushing them is now an error. Wired into the three CLI boundaries that resolve the push decision:

- `tools/package-builder/src/commands/build/all.ts`
- `tools/package-builder/src/commands/build/docker.ts`
- `tools/block-tools/src/cmd/software/build.ts` (bare scenario only; the `target` scenarios push on `location=remote` and never consult `isPrivate`, so they were already immune)

The check runs before the build, not after, so a doomed run does not cross-compile first.

## On the fixture changes

`!isPrivate` entered the autopush default in 7b8fcc427 / #1183 (2025-10-10). At that commit the only private packages carrying `block-software` in this repo were exactly five test fixtures, so `private` was adopted as an implicit "do not push" signal, and at the time it was correct.

What changed since is that the structurer started stamping `private: true` onto every block's software package, so a signal meaning "test fixture" landed on real blocks. This PR migrates those five fixtures onto the explicit `--docker-no-autopush` flag, which is what they always meant. `private` stops being load-bearing for push suppression, and the leftover case can then be an error rather than silence.

## Also

Drops the redundant `&& !core.isPrivate` in `pl-pkg build docker`. The default above already encodes it, and repeating it made the documented `--docker-autopush` opt-in a silent no-op for exactly the packages that most needed it.

## Verification

- 7 new unit tests; package-builder-lib suite 128/128; block-tools 327 passed / 2 skipped
- `turbo run check` clean on all three packages
- End to end against the real private fixture with `CI=true`: the guard fires immediately with the message below, and passes through to the docker build once `--docker-no-autopush` is given

```
error: This CI run would build docker images without pushing them, while still
writing entrypoint descriptors that reference them. Consumers would fail to pull
the images.
  The usual cause is "private": true in package.json: auto-push is gated on
  !isPrivate, and software packages must never be private.
  If this package really must build images without publishing them, say so with
  --docker-no-autopush (or PL_DOCKER_NO_AUTOPUSH=1).
```

The `tests/package-builder` integration fixtures could not be run to completion locally: this machine's docker daemon cannot create veth pairs, so `pip install` inside the image build fails. Environmental and pre-existing, but worth a run on a normal machine.

Ticket: MILAB-6714

<h3>Greptile Summary</h3>

This PR adds a shared CI guard that rejects implicit Docker build-without-push decisions and wires it into the package-builder and block-tools CLI boundaries. It also migrates private test fixtures to an explicit no-push option and permits explicit autopush for private packages in `pl-pkg build docker`.

**Important touched terms**
- **Push intent** — the resolved combination of whether Docker images will be built, pushed, or intentionally left unpublished; this PR introduces `PushIntent` and validates it before builds begin.
- **Docker package** — a `block-software` entrypoint that produces a Docker image; its count is now exposed through the public `Builder` type and supplied to the guard.
- **Autopush** — automatic image publication, normally enabled in CI for non-private packages; this PR rejects implicit suppression and makes explicit `--docker-autopush` effective for private packages in `pl-pkg build docker`.
- **Explicit no-autopush** — `--docker-no-autopush` or `PL_DOCKER_NO_AUTOPUSH=1`; this is now the supported CI opt-out and replaces implicit suppression in five private fixtures.
- **Bare software-build scenario** — block-tools' compatibility mode that derives Docker push behavior from CI/private state; this PR applies the new guard to that scenario.
- **Entrypoint descriptor** — the generated `.sw.json` reference consumed at runtime; the guard prevents CI from silently generating one that points to an image it did not intend to publish.

The PR should not merge until the guard accounts for selected package IDs, otherwise valid non-Docker CI builds can fail in mixed private packages.

The new guard uses a package-wide Docker count while every affected command can restrict actual work with `--package-id`, creating a concrete false-positive failure before the filtered build starts.

**Files Needing Attention:** tools/package-builder/src/commands/build/all.ts, tools/package-builder/src/commands/build/docker.ts, tools/block-tools/src/cmd/software/build.ts
</details>

| Filename | Overview |
|----------|----------|
| lib/node/package-builder-lib/src/core/push-guard.ts | Introduces the shared CI push-intent validation and its diagnostic error. |
| tools/package-builder/src/commands/build/all.ts | Adds pre-build validation, but passes an unfiltered Docker count when package IDs restrict the actual build. |
| tools/package-builder/src/commands/build/docker.ts | Adds the same guard and enables explicit autopush for private packages; selected IDs are likewise omitted from the count. |
| tools/block-tools/src/cmd/software/build.ts | Applies push-intent validation to bare scenarios, with the same package-selection mismatch. |
| lib/node/package-builder-lib/src/core/__tests__/push-guard.test.ts | Covers the guard's boolean branches but not CLI package-ID filtering. |
| lib/node/package-builder-lib/src/index.ts | Exports the guard types and exposes Docker package metadata on the public builder surface. |

</details>

```mermaid
flowchart TD
  CLI[Package-builder or block-tools CLI] --> Resolve[Resolve build and push intent]
  Resolve --> Guard{CI guard}
  Guard -->|Implicit build without push| Error[Fail before compilation]
  Guard -->|Push enabled or explicit opt-out| Build[Build selected artifacts]
  Build --> Push{Push enabled?}
  Push -->|Yes| Registry[Publish Docker images]
  Push -->|No, explicit| Descriptor[Write entrypoint descriptors]
  Registry --> Descriptor
```
</details>

<a href="https://app.greptile.com/ide/claude-code?prompt=%23%23%23%20Issue%201%0Atools%2Fpackage-builder%2Fsrc%2Fcommands%2Fbuild%2Fall.ts%3A86%0A**Guard%20Counts%20Unselected%20Docker%20Packages**%0A%0AWhen%20a%20private%20mixed-artifact%20package%20is%20built%20in%20CI%20with%20%60--package-id%60%20selecting%20only%20non-Docker%20entrypoints%2C%20the%20guard%20still%20uses%20the%20package-wide%20%60core.dockerPackages.size%60%2C%20causing%20the%20valid%20filtered%20build%20to%20fail%20before%20any%20work%20begins.%0A%0A---%0A%0AFor%20each%20issue%20above%2C%20determine%20whether%20it%20is%20valid%20and%20should%20be%20fixed.%20If%20so%2C%20fix%20it%20directly.&repo=milaboratory%2Fplatforma&pr=1779&platform=github"><picture><source media="(prefers-color-scheme: dark)" srcset="https://greptile-static-assets.s3.amazonaws.com/badges/FixAllInClaudeDark.svg?v=6"><source media="(prefers-color-scheme: light)" srcset="https://greptile-static-assets.s3.amazonaws.com/badges/FixAllInClaude.svg?v=6">Fix All in Claude Code</picture></a>

`````markdown
### Issue 1
tools/package-builder/src/commands/build/all.ts:86
**Guard Counts Unselected Docker Packages**

When a private mixed-artifact package is built in CI with `--package-id` selecting only non-Docker entrypoints, the guard still uses the package-wide `core.dockerPackages.size`, causing the valid filtered build to fail before any work begins.

---

`````

</details>

<sub>Reviews (1): Last reviewed commit: ["MILAB-6714: fail CI builds that never pu..."](https://github.com/milaboratory/platforma/commit/d9753309a9d17c2985cb20963a25921d5b3e7174) | [Re-trigger Greptile](https://app.greptile.com/api/retrigger?id=52121983)</sub>

**Context used:**

### DIFF
--- a/.changeset/docker-push-intent-guard.md
+++ b/.changeset/docker-push-intent-guard.md
@@ -1,0 +1,7 @@
+---
+"@platforma-sdk/package-builder-lib": minor
+"@platforma-sdk/package-builder": minor
+"@platforma-sdk/block-tools": minor
+---
+
+Fail CI builds that produce docker images without pushing them. A stray `"private": true` on a software package silently disabled auto-push while the entrypoint descriptor still referenced the image tag, so CI went green and the block 404'd pulling the image at runtime. Opt out with `--docker-no-autopush` when a package really must build images it does not publish.
--- a/lib/node/package-builder-lib/src/core/__tests__/publish-docker-images.test.ts
+++ b/lib/node/package-builder-lib/src/core/__tests__/publish-docker-images.test.ts
@@ -307,3 +307,66 @@ describe("Core.buildDockerImages + publishDockerImages", () => {
     }
   });
 });
+
+// selectedDockerPackages answers "what docker images would this build produce?"
+// for a given --package-id selection. Callers use it to reason about a build
+// before running it, so it must mirror buildDockerImages' own id resolution:
+// a plain intersection with dockerPackages would miss the `:docker` virtual
+// entrypoint that python autogen produces.
+describe("Core.selectedDockerPackages", () => {
+  let tempDir: string;
+
+  beforeEach(() => {
+    vi.clearAllMocks();
+    tempDir = fs.mkdtempSync(path.join(os.tmpdir(), "selected-docker-test-"));
+  });
+
+  afterEach(() => {
+    try {
+      fs.rmSync(tempDir, { recursive: true, force: true });
+    } catch {
+      // ignore
+    }
+  });
+
+  function mixedPackageRoot(): string {
+    return makePackageRoot(
+      tempDir,
+      {
+        name: "@test/mixed",
+        version: "1.0.0",
+        "block-software": {
+          entrypoints: { pyep: pythonEntrypoint, javaep: javaEntrypoint },
+        },
+      },
+      { requirementsTxt: true, runenv: { ref: RUNENV_REF, pythonVersion: "3.12.10" } },
+    );
+  }
+
+  it("counts every docker artifact when nothing is selected", () => {
+    const core = new Core(mockLogger, { packageRoot: mixedPackageRoot() });
+    expect(core.selectedDockerPackages()).toHaveLength(1);
+  });
+
+  it("returns none when the selection has no docker artifact", () => {
+    const core = new Core(mockLogger, { packageRoot: mixedPackageRoot() });
+    expect(core.selectedDockerPackages(["javaep"])).toHaveLength(0);
+  });
+
+  it("resolves the virtual :docker entrypoint for a selected python entrypoint", () => {
+    const core = new Core(mockLogger, { packageRoot: mixedPackageRoot() });
+    expect(core.selectedDockerPackages(["pyep"])).toHaveLength(1);
+  });
+
+  it("returns none for a package with no docker artifacts at all", () => {
+    const root = makePackageRoot(tempDir, {
+      name: "@test/java-only-selection",
+      version: "1.0.0",
+      "block-software": { entrypoints: { main: javaEntrypoint } },
+    });
+
+    const core = new Core(mockLogger, { packageRoot: root });
+    expect(core.selectedDockerPackages()).toHaveLength(0);
+    expect(core.selectedDockerPackages(["main"])).toHaveLength(0);
+  });
+});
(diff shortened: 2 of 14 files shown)


DenKoren (approved):
Does this actually mean we should not disable autopush for private packages unless explicitly set by flags?
We switched from 'all packages are public' model in blocks to 'only block package is public', leaving all internal packages be private.
Maybe, it is a moment to challenge the decision not to push private packages.

PR #1782, opened 2026-08-12
Template pack v4: flat template node map

kevindetry-milaboratories:
## Problem

A block's compiled workflow (`main.plj.gz`) decompresses to far more JSON than
it has content. The published
`@platforma-open/milaboratories.sequence-embeddings@1.4.1` is 821,733 bytes
gzipped and **30,517,238 bytes decompressed**. Only 2.6 % of that is source
code.

`CompiledTemplateV3.hashToSource` deduplicates source *text*, but
`TemplateDataV3.templates` is `Record<string, TemplateDataV3>` — a nested inline
tree. Every reference to a sub-template embeds a full copy of its entire
subtree, so a shared node deep in the graph is written once per path that
reaches it:

```
distinct templates:        39
materialised nodes:    11,672
max depth:                 13

5,836 x ll.get-field
1,804 x workdir.save
1,804 x exec.delayed-compute-allocation.impl
```

87 % of the whole file (26.6 MB) is repeated `libs` metadata records —
`{name, version, sourceHash}` triples re-serialised at every one of those
11,672 nodes. 139,467 reference records point at 139 unique sources.

The compiler already holds a DAG in memory; only the serialised form expands it.

gzip hides this on the wire — duplication this regular compresses 37:1 — so the
existing size guard, which measures gzipped bytes, saw 821 KB against a 3.4 MiB
cap and passed.

## Change

`pl.tengo-template.v4` adds `hashToTemplate`, keyed by a node's content hash;
`templates` holds hash references. Same approach `hashToSource` already used for
source text, applied to the nodes.

Keyed on content rather than `sourceHash` because `hashOverride` exists to give
two templates with the same source distinct identities. Safe because children
are compiled first (topological order) and never mutated once attached — the
compiler only reads them, and `addRequiredCapability` mutates the parent.

## Results

`@platforma-open/milaboratories.sequence-embeddings@1.4.1`:

| | v3 | v4 |
|---|---|---|
| gzipped | 802 KiB | **176 KiB** (4.55x) |
| decompressed | 29.10 MiB | **1.01 MiB** (28.9x) |
| gunzip + `JSON.parse` | 93.3 ms | **3.2 ms** |
| retained heap | 93.1 MiB | **3.1 MiB** |

Across `@platforma-sdk/workflow-tengo`'s own 45 packs, decompressed size drops
from 41.65 MiB to 12.55 MiB.

Both graph walkers get cheaper too:

- `createTemplateV3Tree`'s `updateCacheKey` recursed over the expanded subtree
  on every `createResourceCached` call, including cache hits. For the block
  above that was 35,920 node visits and 2.8M `hash.update` calls to produce 39
  resources. A v4 child's hash already covers its subtree, so the cache key
  folds it in directly and the walk is gone.
- `flattenTemplateTree` memoises by node identity. A v4 pack stores each
  distinct node once, so `resolveTemplate` returns the same object for every
  reference and each node is processed once. A v3 pack has a separate object per
  occurrence, so the memo never hits and behaviour is unchanged.

## Verification

Built `@platforma-sdk/workflow-tengo` 6.8.2 locally under v4 and compared every
pack against the same version published on npm, expanding the v4 result back to
the v3 shape. It canonicalizes byte-identically for 34 of 45 templates. The
other 11 differ in exactly one artifact — `software-ptabler:main`'s `.sw.json` —
and their template graph shape (names, versions, aliases, nesting) is identical.
ptabler is a `workspace:*` dep whose local source has moved past the published
2.1.8 snapshot, so that drift is unrelated.

**0 graph-shape differences. 0 differences outside software descriptors.**

## Compatibility

- **`tengo-builder` now emits v4.** A pack built with this version cannot be
  read by a middle layer older than this release, because the middle layer
  parses `main.plj.gz` itself. This is the rollout constraint to plan around.
- **The backend is unaffected.** Every block path in `block_pack.ts`, including
  `from-registry-v1` and `from-registry-v2`, fetches the pack into
  pl-middle-layer and rewrites it to `{type: "explicit", content}`.
  `loadTemplateFromExplicitDirect` then builds the `TengoTemplateV1` resource
  tree in TypeScript. The backend only ever sees that tree, which is unchanged.
  (`TengoTemplatePack` and `TengoTemplatePackConvert` are declared in
  `template_loading.ts` and referenced by no code path.)
- **v2 and v3 stay readable.** `pl-middle-layer` and `block-tools` read all
  three, so already-published blocks keep working — including their
  `requiredCapabilities` install gate. `required_capabilities.test.ts` guards
  that specifically: reading only v4 would report "no requirements" for every
  pre-v4 block and let a WASM block install on a backend without the runtime.
- `parseTemplate` can now return `CompiledTemplateV4`; the new
  `AnyCompiledTemplate` union covers all three. Exhaustive switches over the
  result need a v4 arm.

## Cycle guard

Readers reject `templates` reference cycles instead of recursing into them. v3
could not express a cycle — it nested children inline, so one would have been an
infinite document — and v4's hash references make it representable.

This is hardening against a corrupt pack, not a live bug: a node's hash covers
its children's hashes, so the compiler would need a hash preimage to emit a
cycle. But readers walk the graph recursively and we do not verify hashes on
read, so the failure mode was a stack overflow rather than a diagnosable error.

## Follow-up, not in this PR

Switching the codec from gzip to zstd takes the same pack from 176 KiB to
128 KiB and halves decompress time. The clean way is a new extension —
`.plj.zst` — which makes the codec self-describing and gives compatibility for
free, since existing files keep their `.plj.gz` name.

Worth stating plainly so it does not get mistaken for a substitute: zstd on the
**unmodified v3** pack reaches 134 KiB, which looks like it solves the problem
and does not. Load time improves 1.1x and heap not at all, because the cost is
`JSON.parse` of 29 MiB, which no codec touches.

<h3>Greptile Summary</h3>

The PR introduces template-pack v4, replacing recursively embedded template nodes with content-hash references into a flat node map while retaining v2/v3 reader compatibility. The previously reported cyclic-reference failure is fixed by tracking the active traversal path in both v4 loading paths.

- Adds `CompiledTemplateV4` and `TemplateNodeV4`, plus hash, root-resolution, reference-resolution, and cycle-error helpers.
- Updates Tengo compilation to emit deduplicated v4 packs.
- Adds v4 support to direct and cached middle-layer template loading, capability extraction, block packaging, serialization, and diagnostics.
- Preserves v2/v3 loading and capability handling for existing packs.
- Adds coverage for cycles, shared nodes, deterministic hashes, capability propagation, and pack-size behavior.
- Important touched terms:
  - **Template pack** — the compressed serialized Tengo workflow consumed by the middle layer; this PR changes newly compiled packs from v3 to v4.
  - **`CompiledTemplateV4`** — the new top-level pack type; it stores source text in `hashToSource`, nodes in `hashToTemplate`, and the root as a hash reference.
  - **`TemplateNodeV4`** — one template graph node containing artifacts, metadata, capability requirements, and child-template hash references instead of nested child objects.
  - **`hashToTemplate`** — the new content-hash-to-node map; it stores each distinct node once and preserves DAG sharing.
  - **`templateNodeHash`** — the canonical SHA-256 identity of a completed node; child hashes make the identity cover the complete subtree.
  - **`AnyCompiledTemplate`** — the shared union of v2, v3, and v4 pack types; consumers now switch exhaustively across all readable formats.
  - **Template cache** — the middle-layer path that flattens packs into topologically ordered cacheable resources; v4 traversal memoizes shared node objects and rejects active-path cycles.
  - **`requiredCapabilities`** — backend feature tokens propagated to the root template and block manifest; readers now retrieve them from either the v3 root object or the v4 root-map entry.
  - **Cycle guard** — active-path tracking using `visiting`; the direct loader and cache flattener now reject cyclic v4 references with a diagnostic error rather than exhausting the stack.

The PR appears safe to merge.

No blocking failure remains.
</details>

| Filename | Overview |
|----------|----------|
| lib/model/backend/src/template_data_v4.ts | Defines the flat v4 pack and node contracts, canonical node hashing, reference resolution, and cycle diagnostics. |
| tools/tengo-builder/src/compiler/compiler.ts | Changes the compiler to seal completed nodes by content hash and emit deduplicated v4 node maps. |
| lib/node/pl-middle-layer/src/mutator/template/direct_template_loader_v4.ts | Renders v4 references into backend resources and now correctly rejects references to nodes active on the current traversal path. |
| lib/node/pl-middle-layer/src/mutator/template/template_cache.ts | Adds v4 graph flattening with identity memoization, topological output, and active-path cycle detection while preserving v2/v3 handling. |
| tools/block-tools/src/v2/build_dist.ts | Extends block-manifest capability extraction to resolve capability declarations from v4 root nodes while retaining v3 compatibility. |
| lib/model/backend/src/serde.ts | Extends the public parsed-template union and discriminator handling to include v4 packs. |

</details>

```mermaid
flowchart LR
  A[Tengo source and dependencies] --> B[tengo-builder]
  B --> C[CompiledTemplateV4]
  C --> D[hashToSource]
  C --> E[hashToTemplate]
  C --> F[Root node hash]
  F --> E
  E -->|Child hash references| E
  C --> G[pl-middle-layer]
  G --> H{Loading path}
  H --> I[Direct resource renderer]
  H --> J[Cached graph flattener]
  I --> K[Active-path cycle guard]
  J --> K
  K --> L[TengoTemplateV1 resource graph]
```
</details>

<sub>Reviews (2): Last reviewed commit: ["Template pack v4: flat template node map"](https://github.com/milaboratory/platforma/commit/e08ddc4e3c312815c3e4df524d91c6b4842b8b84) | [Re-trigger Greptile](https://app.greptile.com/api/retrigger?id=52520586)</sub>

<details><summary><h4>Context used (4)</h4></summary>

</details>

### DIFF
--- a/.changeset/template-pack-v4-flat-node-map.md
+++ b/.changeset/template-pack-v4-flat-node-map.md
@@ -1,0 +1,73 @@
+---
+'@milaboratories/pl-model-backend': major
+'@platforma-sdk/tengo-builder': major
+'@platforma-sdk/workflow-tengo': major
+'@milaboratories/pl-middle-layer': major
+'@platforma-sdk/block-tools': major
+---
+
+Template pack format v4: store the template graph as a flat node map, and
+compress packs with zstd.
+
+`CompiledTemplateV3.template` nests each sub-template inline, so a template
+shared by N parents is serialised once per path that reaches it. Sharing is
+common and the graph is deep, so the expansion is multiplicative. The
+published `sequence-embeddings` block held 39 distinct templates expanded
+into 11,672 nodes; 87 % of its pack was the same `libs` metadata records
+repeated at every node.
+
+`pl.tengo-template.v4` adds `hashToTemplate`, keyed by a node's content hash,
+and `templates` now holds hash references. This is the same trick
+`hashToSource` already used for source text, applied to the nodes.
+
+Packs are also now zstd (level 19) instead of gzip, and are named
+`main.plj.zst`. The extension is the codec: readers pick gzip or zstd from
+the file name, so packs published as `.plj.gz` stay readable indefinitely.
+
+Measured on `@platforma-open/milaboratories.sequence-embeddings@1.4.1`:
+
+| | v3 + gzip | v4 + zstd |
+|---|---|---|
+| on disk | 802 KiB | 128 KiB |
+| decompressed | 29.10 MiB | 1.01 MiB |
+| load (decompress + parse) | 93.3 ms | 2.7 ms |
+| retained heap | 93.1 MiB | 3.1 MiB |
+
+The layout carries the load and memory win; zstd carries most of the
+remaining size win. zstd alone on a v3 pack reaches 134 KiB but improves
+load by only 1.1x and heap not at all, because the cost is `JSON.parse` of
+29 MiB.
+
+Across `@platforma-sdk/workflow-tengo`'s own 45 packs, decompressed size
+drops from 41.65 MiB to 12.55 MiB.
+
+Rendering a v4 pack is also cheaper. A v3 cache key had to walk the expanded
+subtree on every lookup, including cache hits; a v4 child's hash already
+covers its subtree, so the walk is gone. For the same block that was 35,920
+node visits and 2.8M hash updates to produce 39 resources.
+
+Readers reject `templates` reference cycles instead of recursing into them.
+v3 could not express a cycle — it nested children inline, so one would have
+been an infinite document. v4's hash references make it representable in a
+corrupt pack.
+
+Breaking:
+
+- `tengo-builder` now emits v4. A pack built with this version cannot be read
+  by a middle layer older than this release, which parses `main.plj.gz`
+  itself. The backend is unaffected — it only ever sees the unpacked resource
+  tree.
+- `parseTemplate` can now return `CompiledTemplateV4`; use the exported
+  `AnyCompiledTemplate` union. Exhaustive switches over the result need a v4
+  arm.
+- `parseTemplate`, `decompressTemplate` and `deriveRequiredCapabilities` take
+  the codec as a required argument, and `ExplicitTemplate` carries a required
+  `codec` field. Compressed bytes are never inspected to guess how they were
+  written — the caller read them from a path or URL and passes what that name
+  says. `templateCodecForPath` derives it.
+- Building a block emits `main.plj.zst`, and the generated
+  `components.workflow` path in a block's `package.json` changes with it.
+- `@platforma-sdk/tengo-builder` now needs Node >= 22.15 for `node:zlib` zstd.
+- `TemplateDataV3` is unchanged and still parsed. `pl-middle-layer` and
+  `block-tools` read v2, v3 and v4, so already-published blocks keep working
+  and keep their `requiredCapabilities` install gate.
--- a/etc/blocks/blob-url-custom-protocol/block/package.json
+++ b/etc/blocks/blob-url-custom-protocol/block/package.json
@@ -37,7 +37,7 @@
   },
   "block": {
     "components": {
-      "workflow": "@milaboratories/milaboratories.test-blob-url-custom-protocol.workflow/dist/tengo/tpl/main.plj.gz",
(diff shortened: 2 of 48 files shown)


kevindetry-milaboratories:
@greptileai

PR #1783, opened 2026-08-13
MILAB-6705: disk-persist the tree mirror (middle layer)

xnacly:
# MILAB-6705: disk-persist the tree mirror (middle layer)

Implements the middle-layer half of MILAB-6705, specified in
`docs/text/work/projects/disk-persist-tree/spec.md`.

Reopening a project should transfer what changed while the user was away, not the tree
again. On the reference project (7,240 resources, 10.4 MB) a cold open moves 10.6 MB in two
stream rounds; a warm reopen should sit near the 25.8 KB steady-state floor in one round.
No backend change: `constructTreeLoadingRequest` already builds its request from tree state
alone, seeding the non-final frontier and passing everything final as a skip set.

## Derivations from the spec

Four decisions were taken during implementation that refine or contradict what the spec
atoms say. The spec set was deliberately not updated, so they are recorded here.

**1. The session witness is the root's signature bytes, not the session id.**
`work/atoms/150-session-bound-snapshot.md` and `250-snapshot-content.md` say the snapshot
stores the session id, compared on load. It stores the root resource's signature bytes
instead, and compares those against the signature of the freshly resolved root.

A signature is an HMAC over the id, the session and the colour, so byte equality means
every other signature in the file is still live, and inequality means they are all dead.
Same guarantee, and it needs no `GetSessionInfo` RPC: that call exists on `LLPlClient` but
`PlClient` caches only `role` from it, behind a `publicGrants:v1` capability gate, so the
session-id route meant a pl-client change plus a fallback for backends predating the RPC.

It is also strictly stronger. A rotated master secret changes the signatures while the
session id is unchanged, so the session-id key would miss it and leave the case to
`failsafe-cold-retry`. The witness catches it on the key.

**2. Invalidation keys on a build-time stamp, not a package version.**
`350-cache-key-and-eviction.md` keys on the package version. It uses a stamp injected at
build time through rolldown's `define` instead: the git sha when the worktree is clean, plus
the build timestamp when it is dirty.

The atom names its own weakness, that local development can edit the pruning rules without
a version change and hit a stale cache. The dirty-worktree component closes that: iterating
on those rules produces a new stamp on every build. Under `USE_SOURCES=1` the build never
runs, so the constant falls back to a per-process random value and the cache simply always
misses, which is the safe direction.

**3. The stamp lives in pl-middle-layer, not pl-tree.**
The atom's rationale for keying on a version is that it stands in for "the pruning function,
the field filter, the traversal stop rules, the finality predicate". All four are ML's:
`project.ts` passes `projectTreePruning`, `projectTreeFieldFilter` and
`projectTreeTraverseStopRules` into the tree.

A stamp in pl-tree would therefore not invalidate on the changes the atom is actually
worried about. A stamp in ML covers both packages, because `updateInternalDependencies:
"patch"` propagates every pl-tree bump into ML. pl-tree keeps a hand-maintained codec schema
version in the file header, but for a different job: letting a decoder recognise a format it
cannot read. Invalidation is the stamp's; format rejection is the schema version's.

**4. The kill switch is a desktop setting.**
The scope atom asks only that a setting exist as an operational kill switch, default on. It
is wired as a settings toggle following `traversalMode` end to end, rather than an env var.

## Notes for review

**Startup eviction drops other backends' and users' snapshots.** `350-cache-key-and-eviction.md`
says to drop entries whose backend, user, build or schema version is not current, and that is
what it does. The consequence worth being explicit about: someone alternating between a local
backend and a cloud one, or between two accounts, has each connection wipe the other's
snapshots at startup, so those reopens are always cold. Keeping them would mean bounding them
by size alone. The atom's wording is unambiguous, so it is implemented as written.

**An idle warm reopen writes nothing, not once.** `450-acceptance-scenarios.md` asks for
exactly one write on an open-and-idle project. That holds for a cold open. For a warm reopen
where the first refresh changes nothing, the change gate correctly suppresses the write
entirely, since the file on disk already matches the tree. The scenario test asserts at most
one.

`PlTreeState.dumpState()` deliberately reads through an invalidated tree, which the spec
flags as a coincidence rather than a contract. `captureTreeState` now refuses an invalidated
tree via a new `PlTreeState.isValid` getter, so "capture before teardown" is enforced rather
than assumed.

pl-client's lint rules carry an explicit `no-restricted-syntax` guard whose message reads
"resource signatures are ephemeral and change from client session to session (never persist
them)". This change persists them on purpose. The witness and the global-id / signature
split are what make that safe, and the codec goes through the typed
`createSignedResourceId` API rather than casting, so the guard is respected rather than
worked around.

## Testing

Two things the scenario tests forced out, both worth a look:

`hits` was not proof of a warm reopen. It counts a snapshot read and accepted by the witness
check, but the tree can still refuse to apply it, and `loadProjectTree` was reporting
`restored: true` whenever `SynchronizedTreeState.init` did not throw. So the flag could claim
the file on disk described the tree in hand when the open had actually been cold.
`SynchronizedTreeState.wasRestoredFromSnapshot` now reports what really happened, the middle
layer reads that, and a separate `restores` counter sits next to `hits`.

The reopen assertions were checked by breaking `restore()` to return false and confirming they
fail, rather than by trusting that they pass.

The contract test is the one the spec names as first to write and fastest to keep: dump,
encode, decode, restore into a fresh tree, then assert the loading request built from the
restored tree equals the original's, seed for seed and skip for skip. It needs no backend.

Alongside it: finality is proven to be recomputed rather than read from the file (a
predicate that settles nothing yields a tree that skips nothing, from a file that says
otherwise); every truncation point of a valid snapshot is checked to fail cleanly rather
than throw; a damaged payload fails its checksum; an unknown schema version loads as absent;
and a snapshot with a dangling reference is proven to leave the live tree untouched.

<h3>Greptile Summary</h3>

The PR adds a versioned, compressed disk snapshot codec to pl-tree and integrates snapshot restore, periodic/close writes, eviction, observability, and a kill switch into pl-middle-layer. Important touched terms:
- **PersistedTree** — the disk-serializable tree mirror; newly introduced with roots, resource bodies, a signature side table, and a session witness.
- **Session witness** — the root resource’s signature bytes used to reject snapshots from another session; newly checked before payload inflation.
- **SynchronizedTreeState** — the live backend-synchronized tree; now accepts a persisted initial state and reports whether restoration succeeded.
- **TreeSnapshotStore** — the new middle-layer filesystem manager for snapshot reads, atomic writes, misses, eviction, and statistics.
- **Build stamp** — a build-specific cache-key component injected by rolldown; newly invalidates snapshots when middle-layer tree-shaping rules change.
- **Change generation** — the tree mutation counter; newly used to suppress redundant periodic and close-boundary writes.
- **Fail-safe cold retry** — fallback from a failed restored-tree initialization to a normal backend load; newly discards snapshots for authentication, permission, or tree-consistency failures.
- **TreeSnapshotOps** — the new configuration group controlling enablement, write interval, and startup size ceiling.
- **Snapshot schema version** — the on-disk codec compatibility marker; newly rejects byte layouts unknown to the decoder.

The pull request should not merge until startup eviction is restricted to files owned by the snapshot store, because an overridden path can currently cause unrelated file deletion.

Snapshot persistence is broadly fail-safe, but TreeSnapshotStore startup housekeeping applies its destructive removal path to every foreign regular file in a caller-configurable directory; compressed decoding also needs non-blocking resource-exhaustion hardening.

**Files Needing Attention:** lib/node/pl-middle-layer/src/middle_layer/tree_snapshot_store.ts; lib/node/pl-tree/src/persisted_tree.ts
</details>

| Filename | Overview |
|----------|----------|
| lib/node/pl-middle-layer/src/middle_layer/tree_snapshot_store.ts | Adds keyed filesystem persistence and eviction, but startup eviction can delete unrelated regular files from an overridden snapshot directory. |
| lib/node/pl-tree/src/persisted_tree.ts | Adds a carefully bounds-checked binary codec and restore path, although compressed output lacks an application-level expansion limit. |
| lib/node/pl-middle-layer/src/middle_layer/project.ts | Integrates restore, periodic writes, close captures, generation gating, and cold-retry behavior without an accepted blocking defect. |
| lib/node/pl-middle-layer/src/middle_layer/middle_layer.ts | Creates and evicts the store during initialization and drains deferred close writes during shutdown. |
| lib/node/pl-tree/src/synchronized_tree.ts | Adds snapshot restoration state and change-generation exposure to the synchronized tree lifecycle. |
| lib/node/pl-middle-layer/build.node.config.js | Injects a build stamp derived from Git state for snapshot cache invalidation. |
| lib/node/pl-middle-layer/src/middle_layer/ops.ts | Adds default snapshot path and behavior settings, including enablement, write cadence, and size ceiling. |

</details>

```mermaid
flowchart TD
  A[Open project] --> B[TreeSnapshotStore reads keyed file]
  B --> C{Header and witness valid?}
  C -- No --> D[Cold SynchronizedTreeState initialization]
  C -- Yes --> E[Decode and restore PersistedTree]
  E --> F{Restored tree initializes?}
  F -- No --> G[Optional discard and cold retry]
  F -- Yes --> H[Warm incremental backend refresh]
  D --> I[Live project tree]
  G --> I
  H --> I
  I --> J{Generation changed?}
  J -- Yes --> K[Periodic or close-boundary capture]
  K --> L[Atomic snapshot write]
  J -- No --> M[Skip redundant write]
```
</details>

<a href="https://app.greptile.com/ide/claude-code?prompt=Greploop%20milaboratory%2Fplatforma%20PR%20%231783%3A%20work%20through%20Greptile's%20open%20review%20comments%2C%20then%20keep%20reviewing%20and%20fixing%20until%20it%20comes%20back%20clean%20at%205%2F5%20with%20zero%20unresolved%20comments.%0AStart%20by%20reading%20the%20comments%20off%20the%20PR%20itself.%20On%20GitHub%2C%20use%20paginated%20%60gh%20api%20graphql%60%20to%20query%20%60pullRequest.reviewThreads%60%20with%20each%20thread's%20%60isResolved%60%20value%20and%20inline%20%60comments%60.%20%60gh%20pr%20view%20--comments%60%20only%20includes%20conversation%20comments%2C%20so%20do%20not%20use%20it%20as%20the%20findings%20source.%20They%20are%20not%20listed%20in%20this%20prompt%20on%20purpose%3A%20the%20PR%20is%20current%2C%20a%20pasted%20copy%20would%20not%20be.%20Skip%20anything%20already%20resolved%2C%20and%20if%20you%20judge%20a%20comment%20wrong%2C%20say%20so%20rather%20than%20changing%20code%20to%20satisfy%20it.%0A%0APrefer%20the%20Greptile%20CLI%2C%20which%20reviews%20the%20working%20tree%20with%20no%20push%20and%20no%20CI%20run.%20Fall%20back%20to%20PUSH%20LOOP%20only%20where%20a%20step%20below%20says%20to.%0A1.%20Run%20%60command%20-v%20greptile%60.%20Missing%3A%20go%20to%20PUSH%20LOOP%2C%20and%20pitch%20the%20CLI%20at%20the%20end.%0A2.%20Run%20%60greptile%20whoami%60.%20It%20exits%200%20even%20when%20signed%20out%2C%20so%20read%20the%20output%3A%20if%20it%20says%20%22Not%20signed%20in%22%2C%20ask%20the%20user%20to%20run%20%60greptile%20login%60%20%28it%20opens%20a%20browser%2C%20do%20not%20run%20it%20yourself%29%20and%20go%20to%20PUSH%20LOOP%20with%20no%20pitch.%0A3.%20Look%20for%20%60greploop%2FSKILL.md%60%20under%20%60.agents%2Fskills%2F%60%20or%20%60.claude%2Fskills%2F%60%2C%20both%20in%20the%20repo%20root%20and%20in%20the%20user's%20home%20directory.%20Found%3A%20follow%20that%20skill%2C%20never%20pushing%20between%20rounds%2C%20and%20you%20are%20done.%0A4.%20Missing%3A%20run%20%60greptile%20skills%20list%60.%20If%20it%20fails%20as%20an%20unknown%20command%2C%20this%20CLI%20predates%20the%20skill%2C%20so%20tell%20the%20user%20to%20run%20%60greptile%20update%60%20and%20go%20to%20PUSH%20LOOP%20with%20no%20pitch.%20Otherwise%20ask%20the%20user%20before%20running%20%60greptile%20skills%20install%20greploop%60%2C%20then%20follow%20the%20skill%20it%20writes.%0A%0APUSH%20LOOP%20%28only%20when%20a%20step%20above%20sends%20you%20here%29%3A%0A%0AGREPLOOP%20ENABLED%3A%20After%20fixing%20all%20issues%2C%20commit%20and%20push%20your%20changes.%20Greptile%20will%20automatically%20re-review%20the%20PR%20via%20webhook.%20Wait%20for%20the%20review%20to%20complete%2C%20then%20read%20the%20new%20review%20comments%20and%20confidence%20score.%20If%20the%20confidence%20score%20is%20below%205%2F5%20or%20there%20are%20unresolved%20comments%2C%20fix%20those%20issues%20and%20push%20again.%20Repeat%20until%20the%20confidence%20score%20is%205%2F5%20with%20zero%20unresolved%20comments.%20Maximum%205%20iterations.%20Do%20NOT%20open%20a%20new%20PR%20%E2%80%94%20keep%20pushing%20to%20the%20same%20branch.%0A%0AIF%20YOU%20ENDED%20UP%20ON%20PUSH%20LOOP%20BECAUSE%20THE%20CLI%20WAS%20MISSING%3A%20once%20the%20loop%20is%20done%2C%20print%20the%20block%20below%20and%20nothing%20else%20about%20the%20CLI.%20Reproduce%20it%20character%20for%20character%20inside%20a%20fenced%20code%20block%2C%20so%20the%20borders%20line%20up.%20Do%20not%20reword%20it%2C%20translate%20it%2C%20summarize%20it%2C%20add%20to%20it%2C%20or%20drop%20the%20box.%20Print%20it%20once%2C%20and%20do%20not%20bring%20the%20CLI%20up%20again%20later%20in%20the%20session.%0A%0A%E2%95%AD%E2%94%80%E2%94%80%E2%94%80%E2%94%80%E2%94%80%E2%94%80%E2%94%80%E2%94%80%E2%94%80%E2%94%80%E2%94%80%E2%94%80%E2%94%80%E2%94%80%E2%94%80%E2%94%80%E2%94%80%E2%94%80%E2%94%80%E2%94%80%E2%94%80%E2%94%80%E2%94%80%E2%94%80%E2%94%80%E2%94%80%E2%94%80%E2%94%80%E2%94%80%E2%94%80%E2%94%80%E2%94%80%E2%94%80%E2%94%80%E2%94%80%E2%94%80%E2%94%80%E2%94%80%E2%94%80%E2%94%80%E2%94%80%E2%94%80%E2%94%80%E2%94%80%E2%94%80%E2%94%80%E2%94%80%E2%94%80%E2%94%80%E2%94%80%E2%94%80%E2%94%80%E2%94%80%E2%94%80%E2%94%80%E2%94%80%E2%94%80%E2%94%80%E2%94%80%E2%94%80%E2%94%80%E2%94%80%E2%94%80%E2%94%80%E2%94%80%E2%94%80%E2%94%80%E2%94%80%E2%94%80%E2%94%80%E2%94%80%E2%94%80%E2%94%80%E2%94%80%E2%94%80%E2%94%80%E2%95%AE%0A%E2%94%82%20%20You%20can%20run%20greploops%20faster%20locally%20with%20our%20CLI.%20%20%20%20%20%20%20%20%20%20%20%20%20%20%20%20%20%20%20%20%20%20%20%20%E2%94%82%0A%E2%94%82%20%20Install%20it%20at%20https%3A%2F%2Fwww.greptile.com%2Fcli%2C%20or%20I%20can%20install%20it%20for%20you.%20%20%E2%94%82%0A%E2%95%B0%E2%94%80%E2%94%80%E2%94%80%E2%94%80%E2%94%80%E2%94%80%E2%94%80%E2%94%80%E2%94%80%E2%94%80%E2%94%80%E2%94%80%E2%94%80%E2%94%80%E2%94%80%E2%94%80%E2%94%80%E2%94%80%E2%94%80%E2%94%80%E2%94%80%E2%94%80%E2%94%80%E2%94%80%E2%94%80%E2%94%80%E2%94%80%E2%94%80%E2%94%80%E2%94%80%E2%94%80%E2%94%80%E2%94%80%E2%94%80%E2%94%80%E2%94%80%E2%94%80%E2%94%80%E2%94%80%E2%94%80%E2%94%80%E2%94%80%E2%94%80%E2%94%80%E2%94%80%E2%94%80%E2%94%80%E2%94%80%E2%94%80%E2%94%80%E2%94%80%E2%94%80%E2%94%80%E2%94%80%E2%94%80%E2%94%80%E2%94%80%E2%94%80%E2%94%80%E2%94%80%E2%94%80%E2%94%80%E2%94%80%E2%94%80%E2%94%80%E2%94%80%E2%94%80%E2%94%80%E2%94%80%E2%94%80%E2%94%80%E2%94%80%E2%94%80%E2%94%80%E2%94%80%E2%94%80%E2%95%AF%0A%0AIf%20they%20take%20you%20up%20on%20it%2C%20install%20with%20%60npm%20install%20-g%20greptile%60%20%28or%20%60brew%20install%20greptileai%2Ftap%2Fgreptile%60%29%2C%20then%20%60greptile%20skills%20install%20greploop%60.%20Leave%20%60greptile%20login%60%20to%20them%2C%20it%20opens%20a%20browser.&repo=milaboratory%2Fplatforma&pr=1783&platform=github">Fix all with Greploop</a> <a href="https://app.greptile.com/ide/claude-code?prompt=%23%23%23%20Issue%201%0Alib%2Fnode%2Fpl-middle-layer%2Fsrc%2Fmiddle_layer%2Ftree_snapshot_store.ts%3A366-371%0A**Eviction%20deletes%20foreign%20files**%0A%0AIf%20%60treeSnapshotPath%60%20is%20configured%20as%20an%20existing%20or%20shared%20directory%2C%20startup%20eviction%20passes%20every%20regular%20file%20outside%20the%20current%20snapshot%20scope%20to%20%60remove%28%29%60%2C%20including%20files%20this%20component%20did%20not%20create%2C%20causing%20unrelated%20user%20data%20to%20be%20silently%20deleted.%0A%0A%60%60%60suggestion%0A%20%20%20%20%20%20%20%20if%20%28inScope%29%20%7B%0A%20%20%20%20%20%20%20%20%20%20current.push%28%7B%20file%2C%20size%2C%20mtimeMs%20%7D%29%3B%0A%20%20%20%20%20%20%20%20%20%20continue%3B%0A%20%20%20%20%20%20%20%20%7D%0A%0A%20%20%20%20%20%20%20%20if%20%28!name.startsWith%28FILE_PREFIX%29%29%20continue%3B%0A%20%20%20%20%20%20%20%20await%20this.remove%28file%2C%20size%2C%20false%29%3B%0A%60%60%60%0A%0A%23%23%23%20Issue%202%0Alib%2Fnode%2Fpl-tree%2Fsrc%2Fpersisted_tree.ts%3A415%0A**Inflation%20lacks%20an%20output%20bound**%0A%0AA%20damaged%20or%20locally%20replaced%20snapshot%20with%20a%20checksum-consistent%2C%20high-ratio%20compressed%20payload%20is%20fully%20inflated%20before%20structural%20validation%2C%20allowing%20project%20open%20to%20consume%20process-scale%20memory%20and%20become%20unresponsive%20or%20terminate.%20Apply%20an%20application-level%20output%20limit%20appropriate%20for%20the%20maximum%20valid%20snapshot%20size.%0A%0A---%0A%0AFor%20each%20issue%20above%2C%20determine%20whether%20it%20is%20valid%20and%20should%20be%20fixed.%20If%20so%2C%20fix%20it%20directly.&repo=milaboratory%2Fplatforma&pr=1783&platform=github"><picture><source media="(prefers-color-scheme: dark)" srcset="https://greptile-static-assets.s3.amazonaws.com/badges/FixAllInClaudeDark.svg?v=6"><source media="(prefers-color-scheme: light)" srcset="https://greptile-static-assets.s3.amazonaws.com/badges/FixAllInClaude.svg?v=6">Fix All in Claude Code</picture></a>

`````markdown
### Issue 1
lib/node/pl-middle-layer/src/middle_layer/tree_snapshot_store.ts:366-371
**Eviction deletes foreign files**

If `treeSnapshotPath` is configured as an existing or shared directory, startup eviction passes every regular file outside the current snapshot scope to `remove()`, including files this component did not create, causing unrelated user data to be silently deleted.

```suggestion
        if (inScope) {
          current.push({ file, size, mtimeMs });
          continue;
        }

        if (!name.startsWith(FILE_PREFIX)) continue;
        await this.remove(file, size, false);
```

### Issue 2
lib/node/pl-tree/src/persisted_tree.ts:415
**Inflation lacks an output bound**

A damaged or locally replaced snapshot with a checksum-consistent, high-ratio compressed payload is fully inflated before structural validation, allowing project open to consume process-scale memory and become unresponsive or terminate. Apply an application-level output limit appropriate for the maximum valid snapshot size.

---

`````

</details>

<sub>Reviews (1): Last reviewed commit: ["MILAB-6705: do not block a project close..."](https://github.com/milaboratory/platforma/commit/03f9d56c1aeb0bfd11f1025031a78c56f67d729c) | [Re-trigger Greptile](https://app.greptile.com/api/retrigger?id=54381737)</sub>

- Knowledge Base — [pl-tree](https://app.greptile.com/milaboratories/-/custom-context/knowledge-base/milaboratory/platforma/-/docs/pl-tree.md)
</details>

### CODE REVIEW

File: lib/node/pl-middle-layer/src/middle_layer/tree_snapshot_store.ts [hunk 1]
```diff
@@ -0,0 +1,419 @@
...
+      });
+    } catch (e: unknown) {
+      if ((e as NodeJS.ErrnoException | null)?.code === "ENOENT") return;
+      logger.warn(
+        `failed to clear the tree snapshot directory: ${e instanceof Error ? e.message : String(e)}`,
+      );
+    }
+  }
+
+  public get stats(): Readonly<TreeSnapshotStat> {
```
AStaroverov: nit: use function `getStats`

lib/node/pl-middle-layer/src/middle_layer/tree_snapshot_store.test.ts
```diff
@@ -0,0 +1,327 @@
...
+    serverInfo: { instanceId: ops.instanceId ?? "instance-1" },
+    authUser: ops.user === undefined ? "someone@example.com" : ops.user,
+  } as unknown as PlClient;
+}
+
+/** A snapshot with roots and no resources: enough to exercise the store, since the codec is
+ *  tested against real trees in pl-tree. */
+function snapshotFor(root: SignedResourceId): PersistedTree {
+  return {
+    witness: toResourceSignature(Buffer.from(root.split("|")[1], "hex")),
```
AStaroverov: We cannot just a "parse" SignedResourceId here.
Best for me is don't do it at all if it's possible of course.
But if don't have another option move it to `platforma/lib/node/pl-client/src/core/types.ts`

AStaroverov: event more, we already have parseSignedResourceId

AStaroverov (approved):

PR #1816, opened 2026-09-09
feat: add native pframes OOM reproducer block

AStaroverov:
Adds a developer block for reproducing memory exhaustion in the native pframes instance used by Platforma Desktop. Clicking **Run memory stress** joins two N-row inputs into N² rows and sorts the result; the UI requests only the table shape. Input size is configurable, and opening the block does not start the stress workload.

Includes the block packages, workspace registration, lockfile entries, and usage instructions. Completed attempts use a fresh axis domain to avoid cached results. Actual OOM behavior depends on available memory, pframes limits, and spilling.

Validation:
- All five block packages build successfully.
- Type, lint, formatting, Tengo syntax, and block structure checks pass.
- Native pframes smoke check: two 10-row inputs produce 100 rows and 5 columns.
- Full OOM run remains manual; it was not executed during validation.

<h3>Greptile Summary</h3>

Adds a developer-only Platforma block that deliberately constructs and sorts a quadratic native pframes join, allowing Desktop developers to investigate memory-exhaustion behavior. Since the previous review, the table definition was extracted into a reusable builder and covered by a bounded native-driver regression test.

- Adds the block kind, model, UI, empty Tengo workflow, facade package, workspace registration, and usage documentation.
- Requires an explicit **Run memory stress** action; merely opening the page does not request the table shape.
- Validates the configured input size before constructing two N-row columns whose inner join has N² rows.
- Advances `runId` after each attempt so the next table uses a fresh axis domain rather than a cached result.
- Tests the 10×10 join shape, repeated-input identity, and cache-busting identity.

**Important touched terms**
- **Block** — A packaged Platforma extension comprising model, UI, workflow, kind, and facade components. This PR adds the `pframes-oom` developer block and registers all five packages in the workspace.
- **pframes / PFrame** — Platforma’s native tabular computation engine and its column collection. The new block intentionally submits a high-cardinality join and sort to the native engine.
- **PTable definition** — A declarative table query specifying source columns, joins, filters, and sorting. The definition is now isolated in `createOomTableDefinition` so production and regression-test paths share the same query.
- **Inner join** — A join retaining combinations with matching shared-axis values. Two independently keyed N-row columns share a constant group value, producing N² joined records.
- **Axis domain** — Metadata distinguishing the identity/context of a table axis. `runId` is incorporated into the shared axis domain to give completed attempts a fresh table identity.
- **Table shape** — The table’s row and column counts without transferring all expanded records to the browser. The UI requests only this shape to trigger native execution while avoiding browser-side materialization.
- **Driver double** — A test fixture wrapping the real native pframes engine with local blob storage. The new bounded test uses it to verify 100 rows, five columns, and table identity behavior.
- **Tengo workflow** — Platforma’s packaged workflow program. This block includes an intentionally empty workflow because the stress operation runs through the UI pframe service in the Desktop host process.

The PR appears safe to merge; the previously requested core regression coverage is present and no new actionable defects remain.

The manually resolved previous finding is fully addressed by a bounded native-driver test that verifies the quadratic result shape and fresh identity when `runId` changes. The extraction into a shared table builder preserves runtime behavior, and the new test follows established driver setup and disposal patterns.
</details>

| Filename | Overview |
|----------|----------|
| etc/blocks/pframes-oom/model/src/table.ts | Defines the validated quadratic join, sorting workload, and run-specific axis domain shared by the block and test. |
| etc/blocks/pframes-oom/model/src/table.test.ts | Adds bounded native-driver coverage for quadratic shape and cache-busting table identity. |
| etc/blocks/pframes-oom/model/src/index.ts | Connects model data to the extracted pframes table-definition builder. |
| etc/blocks/pframes-oom/ui/src/MainPage.vue | Provides an explicit, guarded stress action, input sizing, warnings, and completion/error status. |
| etc/blocks/pframes-oom/README.md | Documents intended usage, risk boundaries, workload sizes, and regression-test execution. |
| pnpm-workspace.yaml | Registers the five new block packages in the monorepo workspace. |

</details>

```mermaid
sequenceDiagram
  actor Developer
  participant UI as Block UI
  participant Model as Block Model
  participant PF as Native pframes
  Developer->>UI: Select N rows
  Model->>Model: Build two N-row columns
  Developer->>UI: Run memory stress
  UI->>PF: getShape(table handle)
  PF->>PF: Inner join to N² rows
  PF->>PF: Sort both value columns
  PF-->>UI: Shape or resource-exhaustion failure
  UI->>Model: Increment runId
  Model->>Model: Build fresh axis domain
```
</details>

<sub>Reviews (2): Last reviewed commit: ["test: cover pframes OOM join and cache i..."](https://github.com/milaboratory/platforma/commit/09cc4dc7b62438bd2e6b82eda1fa750f46ab3b5a) | [Re-trigger Greptile](https://app.greptile.com/api/retrigger?id=62098937)</sub>


AStaroverov:
@greptileai

AStaroverov:
Pushed a second workload. The block could only pressure native pframes: it asked for the table's shape, so the join stayed inside the engine and never crossed into the middle layer — which left the region the client's crash actually lands in untested.

The value type decides where the payload goes. `Int` columns arrive as typed arrays, outside the V8 heap and unable to exhaust it; `String` columns arrive as one JS string per record and land in the worker's own heap. The new workload fetches every joined record of a String column for that reason.

The join is asymmetric so the inline input stays small — the block model runs in a QuickJS sandbox capped at 8 MB — while the join still multiplies the strings into gigabytes. `text records × characters` is capped at 1 MiB for the same reason: over the cap the model builds no table rather than dying while assembling its own input, and the UI says which value to lower. That cap came from actually hitting `PlQuickJSError: out of memory` at 10000 × 2000.

Measured against a 4 GB heap limit, which is what the desktop middle-layer worker gets on an ordinary machine:

| Text records | Joined rows | Strings | Outcome |
|---|---|---|---|
| 100 | 1 million | 4 GB | survived at 3.86 GB heap, 21 s |
| 200 (default) | 2 million | 8 GB | `FATAL ERROR: Reached heap limit`, 45 s |

The native workload is unchanged and still there — it exercises the other region. Its measurements are now in the README, because they explain why it is hard to kill on purpose: pframes spills to disk, and the run grows slower faster than it grows larger. 100 million rows takes 9 s and 2.9 GB; 900 million exceeds 9 GB and several minutes.

kevindetry-milaboratories (approved):

PR #1825, opened 2026-09-14
fix(exec): include python dependency specs in run environment identity

vgpopov:
## Decisions

**Save the dependency specs into the environment, rather than adding a new identity input.** Conda already saves `env-spec.yaml` and R saves `renv.lock`; python saved `venv` alone and was the only run environment not saving its spec. Mirroring the existing pattern keeps one mechanism instead of two. An earlier draft added a `saveBuilder.dependencies()` method feeding `packagesMap` — discarded once the conda/R precedent was found.

**Iterate dependency names through `maps.getKeys`.** It returns sorted keys, and the order reaches the saved `paths` list. Unsorted map iteration would give a single environment two identities and reintroduce churn.

**Do not restore `softwareRef` in `pyVenvInputs`.** It is commented out (`python/index.lib.tengo:140` and `internal.lib.tengo:7`) and restoring it would also stop two softwares with identical dependencies from sharing one environment — which is a real saving and is safe. Keying on the dependency content fixes the defect without giving that up. The commented-out field remains a latent weakness worth a separate look.

## Why

A python run environment is saved from the workdir holding only the bare venv; dependencies are installed afterwards, into the already-saved directory. The backend hashes that directory at save time (`ctl_notify_run_env.go`, `HashDir`), so the environment's content hash was a function of the interpreter alone.

Two python softwares sharing an interpreter therefore produced identical identities, deduplicated onto one originals-index slot, and every later software resolved to whichever environment was registered first. A script ran under another package's venv and failed importing a dependency it declares; scripts whose imports happened to exist in the borrowed venv passed silently, so the real mis-binding rate is higher than the failure count.

Confirmed against a captured instance. All `…/venv` refs shared one CID; the only split was at a software using a different interpreter build:

| software | interpreter | venv CID |
|---|---|---|
| ptexter | `NG:0x64E` (1.3.22) | `4CD23F35…` |
| ptabler | `NG:0xC08` (1.3.24) | `2E5C3AF9…` |
| per-cell-metrics ×5 | `NG:0xC08` (1.3.24) | `2E5C3AF9…` |

Four of five per-cell-metrics entrypoints resolved to ptabler's environment.

## Test status — please read before merging

**The test in this PR has never been executed.** `block-tools software build` fails in a fresh checkout with `r.selectedDockerPackages is not a function`, after building `block-tools` and `package-builder-lib`. The existing `lib/ptexter` package fails identically, so the cause is an unbuilt toolchain rather than the new fixture. Running it also needs a backend and a `runenv-python` download.

The red phase is therefore owed: the test must be shown failing against `7104497292^` before this fix is trusted. CI is the cheaper place to get that, since the toolchain is already built there.

## Key files to review

1. `sdk/workflow-tengo/src/exec/python/create-venv.tpl.tengo` — the fix
2. `tests/workflow-tengo/src/exec/venv-identity.test.ts` — two entrypoints, one interpreter, differing requirements, asserting different `sys.prefix`
3. `tests/software/venv-identity/package.json` — the fixture; the two entrypoints differ only in their requirements file
4. `.changeset/venv-identity-includes-dependencies.md`

<h3>Greptile Summary</h3>

This PR makes Python virtual-environment identity incorporate the dependency specifications used to construct the environment. It saves those specifications alongside the bare venv before hashing and adds integration coverage for two software entrypoints that share an interpreter but declare different requirements.

- Reconstructs dependency specification files from validated package filenames and adds them to the saved run environment in deterministic, sorted order.
- Extends the Python venv input schema with the mapping between dependency names and package filenames.
- Adds a workspace fixture and workflow test asserting that differing requirements produce distinct virtual-environment paths.
- Adds a patch changeset for `@platforma-sdk/workflow-tengo`.

**Important touched terms**
- **Run environment identity** — The content-derived identity used to register and deduplicate an execution environment. It now includes Python dependency-specification files rather than reflecting only the interpreter and bare venv.
- **Virtual environment (venv)** — An isolated Python installation used to execute software. Its saved representation now carries the specifications whose packages are installed afterward.
- **Dependency specification** — A file such as `requirements.txt` describing Python packages to install. The PR loads these files from the software package and saves them into the run environment.
- **Dependency name** — The descriptor key, such as `requirements`, used by installation logic to select a dependency input. It remains the key in the installation-facing dependency map.
- **Dependency filename** — The filename assigned to a dependency inside the software package. A new `dependencyFileNames` input preserves this mapping and provides safe paths for identity construction.
- **Run-environment saver** — The builder that selects paths and packages included in the saved environment. It now adds dependency filenames in sorted order before producing the environment reference.
- **Deduplication** — Reuse of environments with identical identities. It will continue sharing genuinely identical environments while separating environments whose dependency specifications differ.
- **`sys.prefix`** — Python’s active environment prefix. The new integration test compares it across entrypoints to verify that they resolve to distinct venvs.

The PR appears safe to merge, with the previously reported dependency-key collision addressed and no new actionable defects identified.

The current implementation uses package filenames rather than arbitrary dependency-map keys as saved paths, sorts those paths before constructing saver identity, and retains the dependency-name map expected by installation. The previous finding was manually resolved without explanation.
</details>

| Filename | Overview |
|----------|----------|
| sdk/workflow-tengo/src/exec/python/create-venv.tpl.tengo | Saves dependency specifications under their package filenames and adds them to venv identity in deterministic order. |
| sdk/workflow-tengo/src/exec/python/index.lib.tengo | Passes both dependency references and their corresponding package filenames into venv construction. |
| sdk/workflow-tengo/src/exec/python/internal.lib.tengo | Extends internal venv input validation with the dependency-filename mapping. |
| tests/workflow-tengo/src/exec/venv-identity.test.ts | Adds integration coverage asserting that entrypoints with different dependency specifications receive different venv prefixes. |
| tests/software/venv-identity/package.json | Defines two Python entrypoints sharing an interpreter while differing only in their requirements files. |
| .changeset/venv-identity-includes-dependencies.md | Documents the Python run-environment identity correction as a workflow-tengo patch. |

</details>

```mermaid
flowchart TD
  D[Python software descriptor] --> N[Dependency names and filenames]
  N --> L[Load specification references from software package]
  L --> I[Installation map keyed by dependency name]
  L --> F[Identity map keyed by package filename]
  F --> S[Sort filenames]
  S --> W[Build bare venv and save specification files]
  W --> H[Content-derived run environment identity]
  I --> P[Install dependencies after environment save]
  H --> R[Resolve or deduplicate run environment]
  P --> R
```
</details>

<sub>Reviews (3): Last reviewed commit: ["refactor(exec): pass dependency file nam..."](https://github.com/milaboratory/platforma/commit/e840f9ac7761c49ba821a34380c947ab032bb81a) | [Re-trigger Greptile](https://app.greptile.com/api/retrigger?id=63902847)</sub>


sdk/workflow-tengo/src/exec/python/internal.lib.tengo
```diff
@@ -5,6 +5,10 @@ desc := import(":exec.descriptor")
 pythonVenvInputSchema := {
 	"toolset": "string",
 	"dependencies": {"any": validation.reference},
+
+	// Same references as 'dependencies', keyed by the file name each one carries in the software
+	// package, so the venv saves them under the names their authors gave them.
+	"dependencySpecs": {"any": validation.reference},
```
DenKoren, line 11: nit: why not leaving only one?

tests/workflow-tengo/src/exec/run/venv_identity.tpl.tengo
```diff
@@ -0,0 +1,27 @@
...
+exec := import("@platforma-sdk/workflow-tengo:exec")
+
+self.defineOutputs(["venvA", "venvB"])
+
+swA := assets.importSoftware("@platforma-open/milaboratories.software-venv-identity-test:deps-a")
+swB := assets.importSoftware("@platforma-open/milaboratories.software-venv-identity-test:deps-b")
+
+_whereAmI := func(sw) {
+	return exec.builder().
+		cpu(1).ram("500Mi").
```
DenKoren, line 15: nit: 1/2 of GiB for single println?

tests/software/venv-identity/package.json
```diff
@@ -0,0 +1,58 @@
...
+  "name": "@platforma-open/milaboratories.software-venv-identity-test",
+  "version": "1.0.0",
+  "private": true,
+  "description": "Two python entrypoints sharing one interpreter and differing only in their requirements file. Used to prove each gets its own virtual environment.",
+  "type": "module",
+  "files": [
+    "./dist/**/*"
+  ],
+  "scripts": {
+    "build": "rm -rf ./dist && env PL_BUILD_CHANNEL=dev PL_BUILD_VARIANT=binary PL_BUILD_LOCATION=local block-tools software build"
```
DenKoren, line 11: I afraid this might not work for k8s-based tests: they need docker.
Maybe we have to reflect ptabler build settings (`lib/ptabler/software/package.json`)

DenKoren (approved):

tests/software/venv-identity/package.json
vgpopov: it's another PR

